PR #6, opened 2015-02-19
Init only after dom loaded

anttikissa:
The first thing I tried was something like:

```
<!doctype html>
<script src='awesomplete.js'></script>
<input class=awesomplete data-list='blah, blablah, blablablah, blargh'>
```

This didn't work, though, because awesomplete needed the DOM to be loaded when it initializes.

This PR makes it wait until the DOM has been loaded, so you can put the awesomplete.js directly into `<head>` and it will work just as well as if you put it at the end of `<body>`.

The technique is basically what jQuery does (https://github.com/addyosmani/jquery.parts/blob/master/jquery.documentReady.js), except that it might not work for historical browsers (didn't test on IE either).

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -279,11 +279,19 @@ _.SORT_BYLENGTH = function (a, b) {
 
 function regEscape(s) { return s.replace(/[-\\^$*+?.()|[\]{}]/g, "\\$&"); }
 
-window.document.addEventListener('DOMContentLoaded', function () {
+function init() {
 	$$("input.awesomplete").forEach(function (input) {
 		new Awesomplete(input);
 	});
-});
+}
+
+// DOM already loaded?
+if (document.readyState !== "loading") {
+	init();
+} else {
+	// Wait for it
+	document.addEventListener("DOMContentLoaded", init);
+}
 
 _.$ = $;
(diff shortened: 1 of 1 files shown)


awesomplete.js
```diff
@@ -279,9 +279,19 @@ _.SORT_BYLENGTH = function (a, b) {
...
-	new Awesomplete(input);
-});
+function init() {
+	$$("input.awesomplete").forEach(function (input) {
+		new Awesomplete(input);
+	});
+}
+
+// DOM already loaded?
+if (document.readyState !== 'loading') {
```
arthurvr: the rest of the js uses double quotes.

LeaVerou: What about other states, before `loading`? Are we sure they’re not relevant here?

anttikissa: 1) using double quotes now

2) state of documentReady goes from `loading` to `interactive` to `complete` (https://developer.mozilla.org/en-US/docs/Web/API/document.readyState). DOMContentLoaded fires with the `loading` -> `interactive` transition, so `loading` is the only state when listening to it makes sense.

LeaVerou:
Merged, thanks!

PR #11, opened 2015-02-20
added a live region that announces the selected item

ginader:
This is the one accessibility enhancement that works without any localization efforts.
I would love to also add the announcement that says to use the cursor keys and how many results there are but that would mean to add localization and this would add a huge maintenance overhead.

### DIFF
--- a/awesomplete.css
+++ b/awesomplete.css
@@ -1,4 +1,14 @@
 [hidden] { display: none; }
+[hiddena11y] {
+	border: 0;
+	clip: rect(0 0 0 0);
+	height: 1px;
+	margin: -1px;
+	overflow: hidden;
+	padding: 0;
+	position: absolute;
+	width: 1px;
+}
 
 div.awesomplete {
 	display: inline-block;
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -109,6 +109,14 @@ var _ = self.Awesomplete = function (input, o) {
 		inside: this.container
 	});
 	
+	this.status = $.create("span", {
+		hiddena11y: "",
+		role: "status",
+		"aria-live": "assertive",
+		"aria-relevant": "additions",
+		inside: this.container
+	});
+	
 	// Bind events
 	
 	$.bind(this.input, {
@@ -217,6 +225,7 @@ _.prototype = {
 		
 		if (i > -1 && lis.length > 0) {
 			lis[i].setAttribute("aria-selected", "true");
+			this.setStatus(lis[i].textContent);
 		}
 	},
 	
@@ -265,6 +274,10 @@ _.prototype = {
 		else {
 			this.close();
 		}
+	},
+
+	setStatus: function(message) {
+		this.status.innerHTML = message;
 	}
 };
 


awesomplete.css
```diff
@@ -1,4 +1,14 @@
 [hidden] { display: none; }
+[hiddena11y] {
```
LeaVerou, line 2: `hiddena11y`? Isn’t that nonstandard?
Also, items hidden with `[hidden]` in the script are semantically hidden, i.e. should not be announced to screen readers. AFAIK, `hidden` in HTML5 is the same as `aria-hidden`, and this is how it’s used here.

awesomplete.js
```diff
@@ -109,6 +109,14 @@ var _ = self.Awesomplete = function (input, o) {
 		inside: this.container
 	});
 	
+	this.status = $.create("span", {
+		hiddena11y: "",
```
LeaVerou, line 113: Oh, I see how you’re using it. Wouldn’t a class be more appropriate then? Also, can’t we add these properties to something else? Does it have to be an empty span? E.g. perhaps we could apply these properties to the selected item each time and remove them? Would that work?

PS: If accessibility is this verbose and hard, no wonder why most developers ignore it. I care, and I want to make it accessible, but I doubt the average developer would go through this, and it’s sad. Either the specs or the state of implementations have a serious problem.

LeaVerou:
Hi Dirk,
Thanks so much for this! See inline comments.

awesomplete.css
ginader, line 2: Good point! I was trying to align to the existing conventions as much as possible. I guess I went too far in this case. I like a class much better for that as well :-)

awesomplete.js
ginader, line 113: we _could_ use the existing nodes but that would mean that we would need to give them proper focus to have them announced. At the moment the focus never leaves the input and I quite like that. 
Another option to have the selected option announced is to set the content of the input to the selected option. This works fine but would change the current way it works. We _could_ have that as an option?
Personally I quite like using the aria-live region as it is the easiest and most universal to use. I bigger projects is very convenient to just have one live-region somewhere and just push messages there when you need to announce something (i.e: "you have new message").

ginader:
Thanks for the comments. I will change "hiddena11y" to be a class but propose leave the live-region for now.

awesomplete.js
LeaVerou, line 113: Why? This element doesn’t get focus either. Why can’t we apply whatever this element does to the list item itself? I.e. make each list item an aria-live element when it's selected and remove those attributes when its deselected.

ginader, line 113: The way that a aria-live region works is that it announces its content the moment this content has changed. The list items in the dropdown are generated once and only updated when a new search result is generated. Reading them out then doesn't help. We need to announce the currently selected item so the user can decide to select it. So we can either set focus to the selected item (perfectly valid way to do this but it would change the way awesomeplete works at the moment) or we create 1 hidden live-region and have it read whatever we currently select by pushing the messages there. Sorry Lea there's no other option.

awesomplete.js
```diff
@@ -265,6 +274,10 @@ _.prototype = {
 		else {
 			this.close();
 		}
+	},
+
+	setStatus: function(message) {
+		this.status.innerHTML = message;
```
LeaVerou, line 280: We need to use either `textContent` for botth, or `innerHTML` for both. Getting `textContent` and then setting `innerHTML` could result in angle brackets being parsed as HTML etc.
Also, I disagree with using a separate function just for this one line.

awesomplete.css
```diff
@@ -1,4 +1,14 @@
...
+	border: 0;
+	clip: rect(0 0 0 0);
+	height: 1px;
+	margin: -1px;
+	overflow: hidden;
+	padding: 0;
+	position: absolute;
+	width: 1px;
```
LeaVerou, line 10: Can you link to something where I can read about this hiding method? It’s so verbose... I guess the old `clip(0,0,0,0)` doesn’t work correctly? Why the padding: 0? Why would it have padding?
And we don’t need the IE8 clip. IE8 is not in the browsers Awesomplete supports.

ginader, line 10: It is very verbose indeed. @snookca had done some [extensive research](http://snook.ca/archives/html_and_css/hiding-content-for-accessibility) and that was then adapted in the [HTML5 Boilerplate CSS](https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css#L131) which made this become the de facto standard to hide HTML only visually

awesomplete.js
ginader, line 280: I agree on using `textContent` for both. I think for now it's ok to also get rid of the setStatus helper Method as there is only one place that sends status messages to the live-region. I we decide to maybe add an info status that announces that search results have been found and how many and that the user can use the cursor keys to find them then adding it back would make sense.

PR #14, opened 2015-02-20
Link(s) to license

danbri:
@petetnt - 922c74189e459322599ecbf5a861115633ff7c2f was for the issue you mentioned. I figure linking to the actual version in the repo is useful in case people fork and add complications, although admittedly it isn't ideal having two links.


LeaVerou:
Why have a LICENSE link that doesn’t in any way mention what the license is? Wouldn’t it be better if the text said “MIT License”?

danbri:
Makes sense, yep

LeaVerou:
Actually, “License: MIT license" would be more understandable.
I don’t think we need two links, just linking to the one in the repo should be sufficient…

LeaVerou:
Hey @danbri any chance you could amend this? Or should I close the PR? Thanks!

PR #26, opened 2015-02-20
added a live region that announces the selected item

ginader:
This is the one accessibility enhancement that works without any localization efforts.
I would love to also add the announcement that says to use the cursor keys and how many results there are but that would mean to add localization and this would add a huge maintenance overhead.
Updated: changed hiddena11y property to visually-hidden class

### DIFF
--- a/awesomplete.css
+++ b/awesomplete.css
@@ -2,10 +2,6 @@
 
 .visually-hidden {
 	clip: rect(0, 0, 0, 0);
-	height: 1px;
-	overflow: hidden;
-	position: absolute;
-	width: 1px;
 }
 
 div.awesomplete {
--- a/awesomplete.css
+++ b/awesomplete.css
@@ -1,6 +1,7 @@
 [hidden] { display: none; }
 
 .visually-hidden {
+	position: absolute;
 	clip: rect(0, 0, 0, 0);
 }
 
--- a/awesomplete.css
+++ b/awesomplete.css
@@ -1,5 +1,5 @@
 [hidden] { display: none; }
-[hiddena11y] {
+.visually-hidden {
 	border: 0;
 	clip: rect(0 0 0 0);
 	height: 1px;
--- a/awesomplete.css
+++ b/awesomplete.css
@@ -1,11 +1,9 @@
 [hidden] { display: none; }
+
 .visually-hidden {
-	border: 0;
-	clip: rect(0 0 0 0);
+	clip: rect(0, 0, 0, 0);
 	height: 1px;
-	margin: -1px;
 	overflow: hidden;
-	padding: 0;
 	position: absolute;
 	width: 1px;
 }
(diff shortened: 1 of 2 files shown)


LeaVerou:
Thanks Dirk. Did you see my other message?
Do we have to use an extra element? Won’t it work to just toggle these attributes on the current list item?
(btw you can update PRs, no need for new ones to make a modification)

ginader:
Hi Lea, yes I did see your other message and commented inline here: https://github.com/LeaVerou/awesomplete/pull/11#discussion_r25091213

ginader:
I updated the pull request and answered you questions [here](https://github.com/LeaVerou/awesomplete/pull/11#discussion_r25125604) and [here](https://github.com/LeaVerou/awesomplete/pull/11#discussion_r25125500)

awesomplete.css
```diff
@@ -1,4 +1,14 @@
...
+.visually-hidden {
+	border: 0;
+	clip: rect(0 0 0 0);
+	height: 1px;
+	margin: -1px;
+	overflow: hidden;
+	padding: 0;
+	position: absolute;
+	width: 1px;
+}
```
LeaVerou, line 6: Sorry, I hadn’t realized I was commenting on the wrong PR. 
This element doesn’t get focus, so a lot of the concerns described in the article don’t make sense here.
I don’t want to add current verbosity for potential future changes. 
`padding: 0` and `border:0` aren’t needed in any way. `margin: -1px` also shouldn’t be needed. The clip syntax needs to use commas, this is the IE < 7 clip syntax, and we don’t care about IE < 7.

ginader, line 6: You are right. For this use case some of the properties are not needed. I updated the CSS accordingly.

LeaVerou:
If we're using `clip(0,0,0,0)` then why do we need a width and height or overflow?

ginader:
I haven't been testing using `clip(0,0,0,0)` exclusively across your supported Browsers. Would you say it's supported robustly across the Browsers you're supporting?

LeaVerou:
Sorry, I meant `clip: rect(0,0,0,0)`. Yes, it is supported but not sure if it’s ok with screen readers…

ginader:
From my experience and also [backed here](https://developer.yahoo.com/blogs/ydn/clip-hidden-content-better-accessibility-53456.html) by my former colleagues at Yahoo is that content hidden using `clip: rect(0,0,0,0)` is solidly accessible to assistive technologies. All the other additional methods are to handle behavior in legacy browsers. As you are already not supporting those I'd save we're save to go "light"

LeaVerou:
Um, for `clip` to apply you need `position: absolute` too…

(sorry to keep going back and forth, I’m honestly not trying to avoid merging this; I care about it being accessible very much)

ginader:
I appreciate you making sure your code and UX is as good as possible. The fact that I care about the same thing is the reason why this pull request exists :-)

LeaVerou:
Finally merged, thanks!!

ginader:
Yay! :-)

PR #269, opened 2015-02-22
Reuse Array.prototype.slice & shrink minified size

vlazar:
Makes awesomplete even smaller. Faster as well :smile:

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -7,13 +7,15 @@
  
 (function () {
 
+var slice = Array.prototype.slice;
+
 function $(expr, con) {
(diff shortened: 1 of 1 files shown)


LeaVerou:
If we’re gonna do that, wouldn’t a `toArray` function be better?

vlazar:
Do you mean extract `toArray` function or just rename `slice`? Extracting a function will probably just bloat the code.

LeaVerou:
Yeah, you’re probably right. This is better for gzip too. Merged, thanks!

PR #270, opened 2015-02-22
Pass in global. Makes minified code even smaller.

vlazar:


LeaVerou:
Let’s not get overzealous here. :) We’re merely saving a few bytes in the minified file with this, but it needlessly complicates the code people actually work on.
So, I don’t think I'm gonna merge this, but thanks so much for contributing!

vlazar:
Np, I've just thought the "Ultra lightweight" is kind of the theme for this project :)

I also think people these days are get used to this best practice of passing in dependencies like window or document into function. Maybe I should have named local variable `doc`instead of `document` in this PR for clarity.

vlazar:
Oh, and... Could you please consider adding the minification script to this repo, so anyone could build a minified version? Right now the minified version is outdated and it's misleading.

vlazar:
I've created another pull request which add minification via Closure Compiler Service, which I guess what you use for minification. See #14475

BTW, this PR gives about 1% reduction in minified size, which is not bad at all for this trivial change. Also I don't understand how this can complicate the code.

Can you please reconsider merging this in?

LeaVerou:
Thanks so much re: the other PR!

Regarding this one: 1% in a minified size of around 2.5KB (IIRC) is around …25 bytes. Even in the slowest of connections, 25 bytes won’t really make a difference, and the difference is even smaller or nonexistent in the gzipped version.
The only cases where saving a few bytes really makes a difference is websites the size of Google, which almost always use their own homegrown widgets.
On the contrary, contributors not familiar with this pattern could spend a few seconds wondering why this is there, which IMO is a loss bigger than the 25 bytes of savings.

Passing in a variable is basically a hack, exploiting how minifiers work, which increases both the size and complexity of the development code by a bit, for very little savings. IMO the tradeoff is not worth it, sorry.

vlazar:
More like 40 bytes of 4k, hehe ;)

I agree with your on size difference only makes sense with lot's of libraries (had nice experience optimizing code bloat, so I'm  "a little" biased :)).

I think I now get your point on not bringing a lot of  advanced staff into the codebase (for educational reasons maybe?). However this is not only a minification hack, but good practice of passing arguments into function, instead of relying on global state. Dependency injection, if you will. Also helps with testing.

PR #279, opened 2015-02-22
Use Function.prototype.bind for cleaner code

vlazar:
Also reduces minified size.

As long as following browser compatibility is fine:
Chrome 7, Firefox 4.0, Internet Explorer    9, Opera 11.60, Safari 5.1.4
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind


LeaVerou:
Excellent point, thanks! Can't believe I missed this opportunity to use `bind()`. Btw, it can probably use `this` now instead of `me`.

Issue #16, opened 2015-02-20
List of programming language does not only contain programming languages


prrrnd:
The list I am talking about is [here](http://leaverou.github.io/awesomplete/)
Ruby on Rails, for instance, is a framework. But ruby is a programming language.
I understand that the list is only there for the purpose of the demo, and that this is not a real issue, but I just think it would be good to fix it.

bgammill:
I came here to submit a PR about that. Glad I'm not the only one that noticed that.

arthurvr:
https://github.com/LeaVerou/awesomplete/commit/718adf1d1ffb5b47589eb86bba4574a7f872424a#commitcomment-9820079

prrrnd:
Oh well, if the author doesn't even care.. :rocket:

Flygenring:
Please, people!

It's like discussing if 'Fiat' belongs in a list of cars when it's a brand or that a bike shed can't be colored black because black is not actually a color...

bgammill:
You could argue that ad naseum, but we were just trying to give a little help with consistency. It's a community for a reason. There are pull requests for a reason. There are issues for a reason.

LeaVerou:
/me rolls eyes
This kind of OCD-like behavior also reminds me of [Well, Actually](http://tirania.org/blog/archive/2011/Feb-17.html).
I completely agree with @Flygenring. It’s a demo. It’s purpose is not to teach what programming languages are, but to offer varied examples of strings.
However if it bothers you that much, I wouldn’t object to a PR that changes the “Pick a programming language” label to something that also encompasses frameworks as well. Perhaps “What do you code in?” or something.

schmurfy:
I can understand that you don't care about what is written in your example, fine, but why reject a pull request for it for those it bothers ?

You don't care that's fine but things like having ruby on rails categorized as a language is a common and annoying misconception and the fact you made it might just be a hint.

Your library looks nice anyway ;)

Flygenring:
If the case is that pedantics ruins an example of how to use something (awesome), then I'd suggest to just change the label from "Pick a programming language" to "Pick a computer related term" or even "Pick a sample string" –

Flygenring:
If the case is that pedantics ruins an example of how to use something (awesome), then I'd suggest to just change the label from "Pick a programming language" to "Pick a computer related term" or even "Pick a sample string from the list I chose" – just to make it a catch-all of sorts :)

PR #784, opened 2015-02-23
Ruby on Rails is not a language :)

schmurfy:
Ruby is the language.


arthurvr:
Node isn't too. This would fix issue https://github.com/LeaVerou/awesomplete/issues/16.

LeaVerou:
My reply to #16 applies.

PR #785, opened 2015-02-23
Replace == with ===

cfj:
Three equals everywhere for consistency.


LeaVerou:
`===` has a specific purpose. I don’t see why it should be used in any of these places. The likelihood of `==` producing different results than `===` in any of these cases is exactly zero.

LeaVerou:
Actually, I changed my mind. Merged, thanks!

cfj:
Cool :thumbsup:

PR #1024, opened 2015-02-24
fix for #16

schmurfy:
Since you said you could accept this that way, here is my last attempt ;)


LeaVerou:
O_o
Is this a joke?
The narrowest entity that includes both frameworks and programming languages is “thing”? Seriously?

schmurfy:
partly a joke ;)

PR #1202, opened 2015-02-25
Advanced examples and Autocomplete for email fields #8

sillero:


LeaVerou:
Thanks! Could you please fix the merge conflicts?
Also, please move the JS to index.js. 
Though, ideally, it would be nice to be able to reuse the same code to both display the example and to run it. We can do that by showing the `script` element with CSS, but then it won’t have syntax highlighting. 
Perhaps it would work if we wrapped the script element with `<pre class="lang-javascript"><code>`, hmm...

sillero:
sorry, I screwed things up in this PR, you can delete this one
the right one is #6614

PR #1387, opened 2015-02-25
[BUGFIX] Keydown evts when `ul` hidden shouldn't trigger component handling

alexdiliberto:
This fixes a bug where previously, if the `ul` was hidden, the component would still act upon keydown events for Enter/Esc/Up/Down

To replicate the previous bug: In the demo, you can select `Java` then keyup and down. You know see that the context is not within the input tag iself, however it's within the component which is still selecting items from the previously searched list.

As an aside, it was not apparent what method you were using to minify JS so I just compressed with my own approach.


LeaVerou:
Thanks! Could you please use a getter e.g. `opened` that checks of the popup is open by returning `this.ul && this.ul.hasAttribute("hidden")`? Then you can just check for `me.opened` and other places in the code can use that condition instead of repeating it (same with scripts using the code).

alexdiliberto:
@LeaVerou Sure thing. Added it https://github.com/alexdiliberto/awesomplete/commit/db991e1817f90d7bd5cf2ca1b25d4f4229335bd0

LeaVerou:
Lovely. Now if you could only merge conflicts? Sorry for the delay!

alexdiliberto:
@LeaVerou done

LeaVerou:
Merged, thanks!

This is very minor, but in the future, could you please follow the coding style of the project? (e.g. space after `if`)

PR #9757, opened 2015-02-26
Make awesomplete a CJS/AMD module

drd:
- does the module dance from https://github.com/umdjs/umd
- adds a package.json file, based on @blocka's fork
- will not assume the presence of document, in case you're running this in node

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -7,7 +7,7 @@
 
 (function () {
 
-var _ = self.Awesomplete = function (input, o) {
+var _ = function (input, o) {
 	var me = this;
 
 	// Setup
@@ -348,16 +348,31 @@ function init() {
 	});
 }
 
-// DOM already loaded?
-if (document.readyState !== "loading") {
-	init();
-}
-else {
-	// Wait for it
-	document.addEventListener("DOMContentLoaded", init);
+// Are we in a browser? Check for Document constructor
+if (typeof Document !== 'undefined') {
+	// DOM already loaded?
+	if (document.readyState !== "loading") {
+		init();
+	}
+	else {
+		// Wait for it
+		document.addEventListener("DOMContentLoaded", init);
+	}
 }
 
 _.$ = $;
 _.$$ = $$;
 
-})();
+// Make sure to export Awesomplete on self when in a browser
+if (typeof self !== 'undefined') {
+	self.Awesomplete = _;
+}
+
+// Expose Awesomplete as a CJS module
+if (typeof exports === 'object') {
+	module.exports = _;
+}
+
+return _;
+
+}());
--- a/package.json
+++ b/package.json
@@ -1,0 +1,19 @@
+{
+  "name": "awesomplete",
+  "version": "1.0.0",
+  "description": "http://leaverou.github.io/awesomplete/",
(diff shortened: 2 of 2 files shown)


LeaVerou:
Hi there,

Thanks for contributing!
Please, please, please send separate pull requests per chunk of changes, as now I cannot reject some and accept others. For example, I want to merge the package.json change, but not the UMD one (nothing personal; I just hate UMD; it’s such horribly ugly code) 

Also, why would somebody run this in Node? What's the point of this without a GUI? I’m probably missing something here, but what?

drd:
Hello, 

I'm happy to split up the changes. But, I'm not sure what the package.json will buy you unless awesomplete is at least a CJS module. I don't particularly care about UMD, I just need to be able to `require()` the module using webpack and in Node in my isomorphic React application. Would you accept a change that just sets `module.exports = Awesomplete` and avoids direct use of `document` so that will work for me? Otherwise I'll have to look elsewhere for my auto-completing needs :-)

Thanks!

blocka:
In my fork, I just added a package.json and now when I do `require('awesomplete')` I have access the Awesomplete object on window. So, yes I'm "littering" window a little bit, but it works fine for me.

LeaVerou:
@drd Probably, but keep in mind that `document` isn't only used in the init code. It’s also used to create elements, for the Selectors API and pretty much everywhere, so I worry that adding checks everywhere _might_ end up being too messy and verbose.
Given that Awesomplete isn't really useful without a DOM, could you perhaps explain your needs a bit, in case there's a better solution? E.g. can we convert `document` to a local variable that resolves to something else in your environment? I don’t really use React or webpack, so you need to elaborate a bit more for me to understand how to help :)

afc163:
I need that too.

afc163:
Simply put, using [webpack](http://webpack.github.io/), [browserify](http://browserify.org/) or [spm](http://spmjs.io) for **browser side**, you can easily import js modules via a simple `require('awesomplete')`, only demands module's code shall be CommonJS, just like nodejs.

drd:
@LeaVerou I totally agree — there is no point in using Awesomplete without a DOM. However, if I have a component running on the server that depends on Awesomplete, it should be able to safely `require('awesomeplete')` without a `ReferenceError`. Then, once the same component is mounted in the client browser, all the `document` references will work just fine. So, all I need to do is allow the module to parse and execute top-level code in node. None of the other uses of `document` need to be checked.

LeaVerou:
@drd We can definitely work around that! I guess your solution is probably the best way (checking if document exists).

drd:
@LeaVerou,

I've updated and rebased the pull to only export as a CJS module. I've tested the CJS on our internal application and also as a normal script in this fiddle: http://jsfiddle.net/v8upqxhj/

Thanks!

LeaVerou:
Merci!

LeaVerou:
@drd could you please apply the changes @arthurvr mentioned? Or, even better, @arthurvr could you send a PR?

arthurvr:
@LeaVerou Already done. https://github.com/LeaVerou/awesomplete/commit/47ba3d1edb10760fc231483df60ce337e735868f

PR #12308, opened 2015-02-27
"new" keyword not necessary anymore

mauriciosoares:
Hey there,

When you try to use the plugin without the `new` keyword it would return some errors, this PR menages to fix it.

I still think using the `new` keyword a good idea, but it's not strictly necessary anymore.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -8,6 +8,8 @@
 (function () {
 
 var _ = self.Awesomplete = function (input, o) {
+	if(!(this instanceof Awesomplete)) return new Awesomplete(input, o);
+
 	var me = this;
 	
 	// Setup


LeaVerou:
I’ve thought a lot about this before I released it. I think I’d rather enforce the use of `new` to make it clear that it’s creating an object and not just running a function. I’m not 100% sure though, so I might change my mind and reopen this at any time.

In any case, thanks for contributing!

In future contributions, please follow the code style of the project (braces and newlines even for single statement conditionals)? :)

mauriciosoares:
Sure!

About the code style, since I didn't find any doc about this, I've done it in my way, instead of checking the rest of code, my bad.

PR #12411, opened 2015-02-27
use string as parameter to Awesomplete

mauriciosoares:
Hello Again,

I noticed that the plugin only accepts an instance of the element to properly work, I've added a condition to check if the first parameter is a string, it uses your querySelector helper to return a valid element instance. now the plugin can be used like this:

``` js
new Awesomplete('#inputElement', {list: ['item1', 'item2']})
```

Also, I've cleaned up some trailing white spaces, if that's ok...

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -4,19 +4,19 @@
  * @author Lea Verou http://leaverou.github.io/awesomplete
  * MIT license
  */
- 
+
 (function () {
 
 var _ = self.Awesomplete = function (input, o) {
 	var me = this;
-	
+
 	// Setup
-	
-	this.input = input;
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thanks, I was planning to make this change myself at some point. You don‘t need to check if it’s a string though, the `$` function already does it for you :)

mauriciosoares:
Here it is :)

LeaVerou:
Thanks! Could you please pull and merge conflicts? Sorry for the hassle!

mauriciosoares:
No problem, I should have merged it before,

Now it's ok :D

LeaVerou:
Thanks!

PR #14095, opened 2015-02-28
Remove small duplication

vlazar:


LeaVerou:
Heh, I tried to do that too, but it ended up being too unelegant with the whole prefix stuff. And keep in mind, this is not just for Awesomplete’s code, outside authors might want to use it for scripts built on top of Awesomplete’s API, such as the email domain typeahead in the advanced examples section. In those cases, prefix is not enough. What if they need a suffix too? And so on.
Also, the API of this function is suboptimal: By putting a rarely needed optional argument first, you made users having to type `"",` on every call, pointlessly, over and over again. Optional arguments should be at the end for this very reason. Of course, then it’s not clear any more that it’s a prefix, but it never really was.

Anyway, what I’m saying is thanks so much for contributing, but reducing this duplication needs a bit more thought :)

vlazar:
The reasoning behind this was the my feeling that the `$.regExpEscape` was accidentaly public API. For the private helper those changes kinda looked fine to me. I've tried optional boolean flag as a third argument, but didn't liked it.

Since it's public API, this change of course breaks it. Also, it's bad public API, I completely agree with you.

Keep up good work and don't worry, I'm not offended in any sense by rejected PRs :) It's your "child", your code style and so on.

PR #14475, opened 2015-02-28
Add minification via Google's Closure Compiler Service.

vlazar:
Let's add simple build tool to help keeping minified version in sync with source code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+source = awesomplete.js
+
+banner = "// Awesomplete - Lea Verou - MIT license"
+minified = awesomplete.min.js
+
+minify:
+	@echo "Minifying..."
+	@rm -f $(minified)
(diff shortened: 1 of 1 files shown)


LeaVerou:
This looks super cool and yes, I was using Google Closure Compiler indeed!
So, if I’m understanding this correctly, you just run makefile and it just works, with no other hassle?? :D

vlazar:
Yes, just run `make` and wait :)

Of course it's very simple, but I wasn't sure if you'll want to use Gulp or Grunt, or if Windows support is needed. The same reason I didn't made any changes to README. I'm also non native speaker and prefer to avoid writing documentation by myself, if possible :D

LeaVerou:
Gulp requires Node, so I like this solution better :)
Thanks, merged!

Issue #1030, opened 2015-02-24
Show dropdown on focus


kohenkatz:
Right now, there is no `focus` event bound.  That means that if the user types a few characters and then `blur`s the field, returning to the field does not show the dropdown again.

This may make sense for some uses, but the option to show the dropdown on `focus` is actually useful in two cases:
- Allow resuming selection if `blur`red without needing to type another character.
- Allow setting `data-minchars="0"` to show the options without anything having been typed yet.

I can submit a PR if you're interested.

LeaVerou:
Yes, this was something I considered a lot. In some cases, it makes a lot of sense, but in the general case it would be annoying. The reason it wasn’t included was that I couldn’t decide what the API should be. Allowing `minChars` to be `0` makes a lot of sense.

LeaVerou:
The problem with blur and focus is that they are also triggered when you leave the window and return to it, at least in Chrome :/

kohenkatz:
Well you [already](https://github.com/LeaVerou/awesomplete/blob/gh-pages/awesomplete.js#L118) have it disappearing on `blur`, so I figured that just complementing that with `focus` made sense.

LeaVerou:
What I meant was, originally that’s what I had done, but it resulted in it disappearing and appearing every time I switched to another window. It was quite annoying. So I checked what other autocomplete widgets were doing and they made it disappear for good if the user switched windows.

kohenkatz:
I started playing around with a simple implementation (http://jsfiddle.net/kohenkatz/fmgh29qo/2/), but JS's use of "falsey" values means that the existing code treats a 0 the same as not providing a value.  It just means rewriting the elegant `||` fallthrough to something more explicit.

Maybe the window focus issue could be fixed in combination with the [Page Visibility](http://www.w3.org/TR/page-visibility/) API.  In other words, use only `blur` and `focus` events on the input that don't correspond to `visibilitychange` events. I'm guessing that that'll be a _really_ nasty hack though, so it's probably not a good idea.

fractaledmind:
FWIW, you can easily accomplish this behavior with the `evalutate` method along with setting the `minChars` value to `0`. Using some simple jQuery: 

```
var awesomplete = new Awesomplete(document.getElementById('awesomplete'), {
  minChars: 0
});
$('#awesomplete').on('focus', function() {
  awesomplete.evaluate();
});
```

kritika-singh3:
[...]

PR #14489, opened 2015-02-28
Allow numerical options set to zero

ScottPolhemus:
Adjusts the configuration logic to allow numerical options, such as minchars, to be set to zero (as mentioned in #1030).
- Apply JS option before normalizing values (only if HTML attribute is not present)
- Check if numerical value is specified (not null) before converting to
  integer
- Don’t fall back to initial value when option is set to zero

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -269,7 +269,7 @@ function configure(properties, o) {
 		    attrValue = this.input.getAttribute("data-" + i.toLowerCase());
 
 		if (typeof initial === "number") {
-			this[i] = +attrValue;
+			this[i] = parseInt(attrValue);
 		}
 		else if (initial === false) { // Boolean options must be false by default anyway
 			this[i] = attrValue !== null;
@@ -281,7 +281,9 @@ function configure(properties, o) {
 			this[i] = attrValue;
 		}
 
-		this[i] = this[i] || o[i] || initial;
+		if (!this[i] && this[i] !== 0) {
+			this[i] = (i in o)? o[i] : initial;
+		}
 	}
 }
 


LeaVerou:
Hi there,

Thanks for this PR. However, there are a few issues:
- `null` is not the only falsy value that means "no value provided". For example, `undefined` or `NaN` should behave the same. 
- It’s late and I probably didn’t look at this carefully enough, but I don’t understand why all this complication for a change that only affects numerical values. Why change how function and boolean options work?! And why set `attrValue` to the provided option? 

The way I see it, this change should be doable in ~2 loc that explicitly check for 0 instead of using `||`. But keep in mind it’s 4am here and it’s entirely likely I completely missed something…

ScottPolhemus:
Hi Lea, thanks for taking the time to review my changes (and thanks for this great little plugin!). I probably could have done a better job at explaining the reasoning behind this one.

The problem with an explicit check for zero at the end is that an empty or not present HTML attribute would resolve to zero when you do the `+attrValue` conversion. So we need some check if the attribute has been assigned. I was using `null` for this as that’s what the spec indicates as a return value for getAttribute if the attribute doesn’t exist, however it would be safer to do an explicit check using hasAttribute.

I had moved the JS option assignment to the top of the function as a way to normalize the values the same way regardless of how they are set. For instance, this would enable numerical options being set using a string (e.g. `minChars: "1"`). Looking at it again I agree it might have overcomplicated things.

I took another stab at it which is hopefully a bit more clear in my intentions.
- Initialize property to `null` to simplify fallback logic
- If the HTML attribute is present (explicit check using `hasAttribute`), use the existing if/else statements to parse and assign a value.
  - Check `isNaN()` before applying the value of a numerical option (to avoid an empty string resolving to zero)
- If `this[i]` is still `null`, that means the attribute doesn’t exist, the initial value is a function, or a nonnumerical value was provided for a numerical option.
  - Therefore we should check for a JS option, and fall back to this initial value if there is none.

LeaVerou:
> The problem with an explicit check for zero at the end is that an empty or not present HTML attribute would resolve to zero when you do the +attrValue conversion. So we need some check if the attribute has been assigned.

Or we just change the way the conversion is done instead of adding six trillion lines of code to work around it.

>  safer to do an explicit check using hasAttribute.

safer how?! null means that the attribute wasn't provided and NOTHING ELSE. If it was provided then it will be a string, even if it's an empty string. 

> If the HTML attribute is present (explicit check using hasAttribute), use the existing if/else statements to parse and assign a value.

Why do this check? In non-boolean options, there's no difference between not providing the attribute at all, or providing an invalid value (e.g. an empty string instead of a number). 

Please don't change the way the function works just because it is not to your liking. I understand it’s not how you would have written it, and I hate other people’s code as much as the next programmer :P However, if you have objections to the actual logic here, please send a separate PR after this one is merged justifying your view. This PR is about allowing 0 for numerical options and nothing more. A reasonable PR for this only needs to make changes in two places: The `this[i] = +attrValue;` line and the `this[i] = this[i] || o[i] || initial;` one. I’m very tempted to just go in and do it myself, but since it was your idea, I think it should be you in the commit log that added it.

ScottPolhemus:
Alright, let's try this again. ;)

Here’s a new version with just the changes you suggested. Numerical options are only applied if the attribute is a number, and it falls back to the JS or initial value if the result is falsey and not zero. If you’ve got an even simpler solution in mind feel free to make the edit yourself. I don’t care much about the credit, but I appreciate you giving me the chance to get it right. (FWIW this is the first open source project I’ve even made any contributions to.)

The "safer" comment was in reference to the fact that some DOM implementations might follow the older spec and return an empty string instead of `null` (source: https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#Notes). `isNaN()` will catch both cases, though, so no reason to worry about that.

awesomplete.js
```diff
@@ -281,7 +281,9 @@ function configure(properties, o) {
 			this[i] = attrValue;
 		}
 
-		this[i] = this[i] || o[i] || initial;
+		if (!this[i] && this[i] !== 0) {
+			this[i] = typeof o[i] !== "undefined"? o[i] : initial;
```
LeaVerou: I'd prefer `!(i in o)` which shorter and more clear. Is there any specific reason you did it this way?

LeaVerou:
Thanks, this is much better! I only have one small comment, which I've added inline.

Thanks for not being put off by criticism, if I knew it was your first OSS contribution I would be more mindful of how I expressed it. Sorry for that!

Regarding the older spec, Awesomplete doesn't support older browsers anyway and AFAIK all modern browsers return `null`. If you know of any modern browsers that don’t, please let me know.

Thanks again!

awesomplete.js
ScottPolhemus: Nope, you're right, using `in` makes perfect sense here.

LeaVerou:
Oops, didn’t realize you made the change since you didn’t post (I don’t get notified for commits). Thanks, merged!

PR #14490, opened 2015-03-01
Add an event for when the highlighted item changes

ScottPolhemus:
This adds an event called `awesomplete-move` which fires at the end of the `goto()` method.

I needed this because I'm displaying a long list of options in a scrollable container, and I wanted to allow the user to scroll through the list using the arrow keys. With this event I can update the scroll position to ensure the highlighted item is visible.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -176,6 +176,8 @@ _.prototype = {
 			lis[i].setAttribute("aria-selected", "true");
 			this.status.textContent = lis[i].textContent;
 		}
+
+		$.fire(this.input, "awesomplete-highlight");
 	},
 
 	select: function (selected) {
--- a/index.html
+++ b/index.html
@@ -271,6 +271,11 @@ awesomplete.list = ["Ada", "Java", "JavaScript", "Brainfuck", "LOLCODE", "Node.j
 				<td>The popup just closed.</td>
 				<td>No</td>
 			</tr>
+			<tr>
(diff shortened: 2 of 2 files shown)


LeaVerou:
Awesome, thanks! No objections (and love how you also added documentation!), except the naming. As a user, iIf I hadn’t looked at the docs, I would have no idea what `move` does and I would probably assume something entirely different.
I would suggest `change` but I’m not too fond of it and `selectionchange` is too long. Anyone has a better idea for the name? What do other autocompletes use for this event? jQuery UI doesn’t seem to have a smiliar event and Twitter Typeahead uses `cursorchanged` which is a spectacularly terrible name (cursor?!).

drd:
How about `awesomplete-highlight` ?

ScottPolhemus:
@drd I was about to suggest the same thing. My first instinct was "change" as well but I wanted to avoid confusion with the input's native `change` event or the `awesomplete-select` event. `awesomplete-highlight` seems like a good way to indicate an item has been highlighted but not yet selected.

LeaVerou:
Yes, I like it! Can you make the change please? Thanks!

ScottPolhemus:
Gladly. Just pushed a rebased version to this branch with the name changed to "highlight".

Saratpattnaik12:
Good morning everybody

LeaVerou:
Merci!

PR #14564, opened 2015-03-02
UX: Select first option when the option only one

temon:
It's my idea for better UX. 

When we hit Enter button and showing only one option then it will select the option.
Because when people see its only one option showing and he hit enter, means he want to select the option.

thanks

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -64,7 +64,7 @@ var _ = self.Awesomplete = function (input, o) {
 		"keydown": function(evt) {
 			var c = evt.keyCode;
 			
-			if (c === 13 && me.selected) { // Enter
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thanks for contributing!
However, replacing my (=user) input with a suggestion when nothing is selected is terrible UX, sorry. I might be pressing Enter to submit the form, and then get completely unexpected behavior.
You can enable the `autoFirst` option to automatically select the first option though, which should behave in the way you want. I would advise using it only when the likelihood of the user wanting to use one of the suggestions instead of freeform text is > 50%.

temon:
ok @LeaVerou no problem, its my bad. I not read the autoFirst option first.

thanks.

PR #15170, opened 2015-03-03
Added tab functionality

lindekaer:
Now you can select an item by hitting tab (in addition to enter)

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -64,7 +64,7 @@ var _ = self.Awesomplete = function (input, o) {
 		"keydown": function(evt) {
 			var c = evt.keyCode;
 
-			if (c === 13 && me.selected) { // Enter
+			if ((c === 13 || c === 9) && me.selected) { // Enter
 				evt.preventDefault();
 				me.select();
 			}


LeaVerou:
Hmmm, not sure that’s a good idea usability wise. Perhaps not if combined with autofirst. What do others think?

lindekaer:
I've encountered many users who found it natural to tab select; I too find it a pleasant addition, but ultimately a matter of taste.

LeaVerou:
I can see how it could be desirable when the user has already interacted with the autocomplete suggestions. However, in the general case, if `autoFirst` is true, the first suggestion will be selected automatically, whether the user actively engaged with the widget. In that case, tabbing doesn’t necessarily mean "accept suggestion", it could mean "move to the next element"…
Perhaps we could check if autoFirst is false too? But then it might lead to inconsistent behavior…

ScottPolhemus:
I could see this being useful, but probably not a good idea as the default. e.g. In a situation where the plugin was extended to allow multi-option selection, I would definitely expect tab to select the current item. However having tab submit by default would break my current use-case, which combines `autoFirst` and `minChars: 0` to show the options list immediately on focus, placed next to with some other input elements that should be tab-focusable.

My preference would be to hide this feature behind an option (off by default) or leave it out completely. Maybe it's worth adding an advanced example that does this with a separate event listener?

lindekaer:
Lea, I see your point with the autofirst issue. As you propose, I think a great workaround would be detect whether or not the user has actively engaged with the list.

PR #16435, opened 2015-03-10
Remove native autocomplete

goulvench:
Because it overlays the ul while being unreachable using the keyboard.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -13,6 +13,7 @@ var _ = function (input, o) {
 	// Setup
 
 	this.input = $(input);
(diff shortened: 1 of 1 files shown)


ScottPolhemus:
+1 for this.

Seems like most other autocomplete plugins do this, and I have run into the conflicting behavior in practice as well. (It's not too difficult to set the attribute manually, but I can't think of a scenario where you would want both Awesomplete and the browser's native autocomplete functionality.)

LeaVerou:
Merged, thanks!

PR #16556, opened 2015-03-18
New Design

yxuko:
I had this idea to make a better design

### DIFF
--- a/awesomplete.css
+++ b/awesomplete.css
@@ -69,7 +69,7 @@ div.awesomplete > ul:empty {
 
 	div.awesomplete > ul > li {
 		position: relative;
-		padding: .2em .5em;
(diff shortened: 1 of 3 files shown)


LeaVerou:
Screenshot?

yxuko:
capture d ecran 1
capture d ecran 3
capture d ecran 4

LeaVerou:
There is definitely some potential there (and thanks for the effort!), but as it currently stands I find the strokes (borders) way too conspicuous and distracting and the text in this screenshot is effectively unreadable. 

In any case, I’m not quite sure why this is _better_, it’s just a completely different design. I would welcome some critique on the existing one much more. I’m sure there is a lot that can be done on the current retro-ish design to improve it without having to switch to an entirely different aesthetic.

PR #16639, opened 2015-03-28
Match input text against the textContent of options, not innerHTML.

eggpi:
Certain characters in the `innerHTML` property are escaped, and we don't want to match against their escape sequences. For example, try searching for "AT&" or even "amp" in http://jsfiddle.net/vfpf4vor/.


LeaVerou:
Thanks!

I agree on changing this to textContent but it needs to be changed in all places then, not just there.

eggpi:
Sorry, I don't see what else needs to be changed here. There are only two other occurrences of `innerHTML` in awesomplete.js, but both seem unrelated to this issue. Did I miss anything?

eggpi:
Hi @LeaVerou, it's been a few weeks since I opened this PR. Can you please take another look at this when you have the time? I'm really not sure what else needs changing here. Thank you!

LeaVerou:
Sorry for the delay. I merged it anyway, could you please keep an eye on the issues in the next few days in case there are any bug reports related to this? Thanks!

eggpi:
Sure, thanks!

Issue #14565, opened 2015-03-02
Proposal: Tests


vlazar:
Do you have plans to cover code with tests?

It seems reasonable to introduce tests for such a popular project ;)

If you are interested, I'd like to know your preferences in test/mock framework, test runner, etc. Continuous Integration would be also nice to have, as well as versioning and other staff.

But let's just start small. So... tests, YES or NO?

LeaVerou:
Tests would be nice, don’t really have a preferred framework, I didn’t like anything I’ve tried. :/ 
In general, the simpler and easier for new contributors, the better.

arthurvr:
Sure! :+1:

michaelachrisco:
Ive used http://qunitjs.com/ in the past. Its pretty straight forward. We could (all)? do some collab sometime. At the very least, it will enable future pull requests to be tested.

LeaVerou:
Oh I forgot to say: I would prefer something as popular as possible, as it increases chances that contributors will already be familiar with it.

arthurvr:
I'm personally +1 to QUnit.

vlazar:
Here's some popular options sorted by number of stars on github:

Testing Frameworks
https://github.com/jasmine/jasmine
https://github.com/mochajs/mocha maybe with https://github.com/chaijs/chai
https://github.com/jquery/qunit

Test Runners
https://github.com/karma-runner/karma
https://github.com/airportyh/testem

LeaVerou:
I haven’t used them extensively, so whatever the majority goes with, I’m fine with it.
Ideally something which makes it super easy for new contributors to set it up and/or write new tests.

juanto121:
While trying jasmine framework, which I found to have nicer syntax than qUnit and not to forget its ease of use, wondered into the problem of testing the Awesomplete private functions, are they testable? 
I couldnt come up with any way of easily testing them.

vlazar:
So test framework is selected. Excellent!

Here is a better way to run tests with Karma test runner #16734
It would be so much easier to do TDD/BDD and we can later run tests on Travis CI.

juanto121:
There's still discussion about the merged framework #16640. I tried to refactor the code style a bit in feace11ae43a631703dfa59a4a3276b2914c60ae but we should all agree on keeping Jasmine(or changing it) before adding coverage and automation.

vlazar:
Ok, I think it will be easier to say if the tests are easy or not, when you see them.

[...]

PR #16640, opened 2015-04-01
Jasmine simple testing

juanto121:
Sample for awesomplete test using Jasmine. 
Didn't receive feedback on #14565 so I decided to PR. What do you think?

### DIFF
--- a/tests/jasmine/SpecRunner.html
+++ b/tests/jasmine/SpecRunner.html
@@ -1,0 +1,24 @@
+<!DOCTYPE html>
+<html>
+<head>
+  
+  <meta charset="utf-8">
+  <title>Jasmine Spec Runner v2.2.0</title>
+
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/jasmine/2.2.1/jasmine.css">
+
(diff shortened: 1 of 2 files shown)


juanto121:
Hi @LeaVerou , I know you must be really busy but any input on the matter would be much appreciated. Maybe Im missing something obvious?

LeaVerou:
Oh my, it’s been 8 days already since this? So sorry!!

As someone who has never really used Jasmine, I have no idea how to write a test based on the sample code. I could of course go RTFM but I guess contributors will also have the same issue. Perhaps you could add a little quick intro on that? (perhaps in the README?)

juanto121:
Ok, I'll work on it.

LeaVerou:
Hey @juanto121, I went ahead and merged it anyway. Thanks for the work! But it would be nice if you could add some short documentation (referring to the original Jasmine docs where necessary). Also, could you please rename SpecRunner.html to index.html so that people don't have to type a filename as well?

vlazar:
@LeaVerou Did you compared real world examples of tests in Jasmine vs QUnit or just 1 small test which tests that true is true? The difference in later case is not so obvious I think.

PR #16715, opened 2015-04-18
Implement additional event for click on autocompletion list - fix #16712

### LINKED ISSUE #16712: Touch on completion list not firing on iOS
Hey Lea,

I just put awesomplete in a site I'm developing atm. Unfortunately touch on the auto completion list is not filling the input field.

Wouldn't make changing the `mousedown` handler to "normal" `click` do the job also?

```
    // change this to click
    $.bind(this.ul, {"mousedown": function(evt) {
        var li = evt.target;

        if (li !== this) {

            while (li && !/li/i.test(li.nodeName)) {
                li = li.parentNode;
            }

            if (li) {
                me.select(li);
            }
        }
    }});
```

Thanks for the nice project.

stefanjudis:
So second try.

I changed the event handling for "click" on the completion list to be `mousedown` and `click`.
The event handler is wrapped by a debounce function to avoid calling the handler multiple time.

Tested in:
- Chrome
- Firefox
- Safari
- IE 11
- iOS emulator


LeaVerou:
If this is done to fix this on touch devices, why can’t we just add a touch event?

stefanjudis:
> If this is done to fix this on touch devices, why can’t we just add a touch event?

It's working fine on Android at the moment - only problem is mobile safari.

So you'd go with binding `mousedown` and `touchstart` on the `ul`?

But touch is not that good supported isn't it? And now that there is the Pointer Events vs. Touch Events discussion going on, I'm not sure, if mouse and touch are the right way.

http://caniuse.com/#feat=touch

As `mousedown` is a hack for safari and unfortunately is not working on mobile safari I just thought of going the safe road with `debounce`. 

`click` is fired for everybody except desktop Safari, where `mousedown` comes into play.

erquhart:
`click` does seem like the way to go here. @LeaVerou was there a reason for choosing `mousedown` instead?

LeaVerou:
The reason for mousedown was that Safari was failing with click, though I don't remember the why.

LeaVerou:
According to [this](http://sitr.us/2011/07/28/how-mobile-safari-emulates-mouse-events.html) it seems from a quick read that `mousedown` should be fired on Mobile Safari, as long as there is a `onclick` handler on the element (even if empty). Have you tried that @stefanjudis? What about other mobile browsers?

stefanjudis:
> Have you tried that @stefanjudis? What about other mobile browsers?

If I remember correctly I tried the empty eventlistener and it didn't work.

I shipped with a hacked version of `awesomplete` - so feel free to close this one as no other is complaining about this issue. :)

MHanslo:
@stefanjudis, could we consider getting this PR on track again and resolve the conflicts? I shipped a release with a hacked version of awesomplete as well that included a similar fix you wrote here.
Shout if you need help.

stefanjudis:
@MHanslo I don't have the time for that right now and will close it. But feel free to open a new one and adjust the changes. :) Sorry.

PR #16716, opened 2015-04-19
Intro to testing on jasmine

juanto121:
I think I cover the most basic blocks of Jasmine testing, should I be more detailed? should I give an example with an awesomplete functionality?. Let me know!


README.md
```diff
@@ -1,4 +1,4 @@
-# Awesomplete
+|# Awesomplete
```
LeaVerou, line 1: Hi, this broke the heading, could you please fix it? Thanks!

LeaVerou:
Merged, thanks! Could you please fix the readme heading?
I merged it because any documentation is better than none, but it’s still quite unclear how to test any UI interaction. I guess Jasmine can’t help with that?

arthurvr:
> I merged it because any documentation is better than none, but it’s still quite unclear how to test any UI interaction. I guess Jasmine can’t help with that?

You really shouldn't be documenting the test framework you're using in your project's readme (for sure not when your test framework is one of the most-used ones). Contributing sections of a readme should say people to write tests before submitting something, that's true, but the readme shouldn't be teaching jasmine to people. It's not specific, hard to maintain and for 90% of the people reading this a waste of time.

PR #16736, opened 2015-05-16
Update index.html

gabrielduque:


LeaVerou:
It’s minified & gzipped size. I just verified with Closure Compiler and it’s 1.83KB.

PR #16738, opened 2015-05-20
limit saturation to allowed 100% value

ste-m:
Hi,
saturation value in HSL should be in the range 0% - 100%.
Having them over 100% also prevents successful compilation when including awesomplete.css in libsass projects.


LeaVerou:
Oops, no idea how that happened. Merged, thanks!

PR #16734, opened 2015-05-08
Karma test runner

vlazar:
Replaces plain Jasmine spec runner with [Karma](http://karma-runner.github.io/) test runner.

Karma test runner is a much more convenient way to run tests. It allows to run tests in multiple browsers locally or in the cloud, shows failures in console, provides debugger, can be used with Continuous Integration server as Travis CI, and so on and so forth.

PR #16740, opened 2015-05-29
Fix for open() will not open list before any input event

steelywing:
Reproduce:
- Don't fire any `input` event (don't key in anything)
- [Optional] Set input via JS `$('input').val('test')`
- Call `awesomplete.open()`

Issue:
- List will not open

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -107,6 +107,8 @@ var _ = function (input, o) {
 	else {
 		this.list = this.input.getAttribute("data-list") || o.list || [];
 	}
+	
+	this.evaluate();
 
 	_.all.push(this);
 };


vlazar:
Nice catch!

@LeaVerou Any chance you'll have some free time in the near future to look at #16734 ? Fixing bugs would be safer if we have quick and easy way to run tests and, of course, more tests (which I already have and waiting for you response on #16734).

fractaledmind:
This code, in my testing, will have the dropdown appear on page load if `minChars` is set to `0`. I doubt this is desirable behavior, so a more nuanced solution is probably needed.

LeaVerou:
I would welcome that more nuanced solution :)
Closing this PR for now, since that’s a pretty major bug. Feel free to reopen if you want to extend it!
Thanks for contributing steelywing!

LeaVerou:
Also, an easy temporary solution to anyone with this problem would be to fire a synthesized input event on their input.

vlazar:
@LeaVerou What is desired behavior on open(), on input focus and on page load when `minChars` is set to `0`? See related https://github.com/LeaVerou/awesomplete/issues/16761#issue-103575454

PR #16743, opened 2015-06-22
allow users to create their own wrappers

erquhart:
This fixes #16718. Some layouts require the targeted input to remain siblings with other elements - this PR accommodates through the following changes:
- allow the user to create their own wrapper div via the `awesomplete-wrapper` class
- if an input bearing the `awesomplete` class does not have such a wrapper at runtime, Awesomplete will make one just as before, albeit with the `awesomplete-wrapper` class for consistency/simplicity
- CSS selectors updated accordingly
- docs updated as well (hopefully they work, you'll want to test)

My apologies for not writing tests for this, didn't have time to get to it.

PR #16745, opened 2015-06-29
Allow container to be placed on a different object

jbellsey:
This commit allows the user to force the awesomplete container (and its elements) to be placed around an element other than the provided input.

Why? As it stands, Awesomplete is incompatible with the Insignia tag-editing library (https://github.com/bevacqua/insignia). That library requires that the input field has no siblings, and Awesomplete creates siblings.

I have no doubt that this small change will be useful for other installations as well.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -41,7 +41,7 @@ var _ = function (input, o) {
 
 	this.container = $.create("div", {
 		className: "awesomplete",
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thanks for working on this.

I think this issue can be fixed by #16743, no?

PR #16746, opened 2015-06-30
Autocomplete false, and new _() instead of new Awesomplete()

### LINKED ISSUE #16744: Awesomplete is not defined
I have a use case where at the time `init()` is fired, Awesomplete doesn't exist yet.
This is because the DOM exists and is ready at the time the library is loaded.

image

It seems you have attempted to address this already here:

```
// DOM already loaded?
if (document.readyState !== "loading") {
    init();
}
```

However this code precedes the Awesomplete assignment.

```
function init() {
    $$("input.awesomplete").forEach(function (input) {
        new _(input);
    });
}
```

The above will fix it if you think that's acceptable? Let me know and I will put in a pull request

marc-costello:
`autocomplete="off"` no longer works in chrome.
http://stackoverflow.com/questions/15738259/disabling-chrome-autofill

The second commit fixes https://github.com/LeaVerou/awesomplete/issues/16744


LeaVerou:
Merged, though initialization code should probably be after `Awesomplete` is exported to `self`, but that's another issue. 
Thanks!

LeaVerou:
In the future, please try to open 1 PR per issue you're fixing. If I only wanted to merge 1 fix here, I wouldn't be able to.

marc-costello:
@mc808 Yes "off" is the correct standard, the problem we have is Chrome doesn't seem to be follow it, at least not for all versions. As many people suggest in the SO link above - "false" fixed it for me. It is indeed bizarre that "off" works for you on Chrome 44/Windows and it does not for me.

marc-costello:
This (https://code.google.com/p/chromium/issues/detail?id=468153) see's quite a heated discussion on the matter with the Chromium team. There maybe a better solution out of this, albeit not a nice one.

mc808:
Oops I was commenting on the page for the commit itself, which didn't show this back story.

Looks like there are two different auto-thingies in conflict with each other: autocomplete (showing a dropdown list of previously typed inputs) and autofill (automatically filling in credentials when a login form is detected). If I'm reading right, setting "off" will disable the autocomplete dropdown that was plaguing me, while setting any invalid value such as "false" will disable autofill but won't disable autocomplete... 

I hope that's not the case. Otherwise I wonder if it would work to set "off" on the input element and "false" on the form (assuming there's a form)?

lydell:
For me, using the latest Firefox and Chrome versions (on Ubuntu) I get the native dropdown of previously entered stuff on top of awesomplete’s dropdown with suggestions. If I change to `autocomplete="off"` (`off` instead of `false`) that native dropdown disappears, which is what I want.

As far as I understand, `autocomplete="off"` was added to awesomplete for one reason: To disable the native dropdown. The linked SO question in the OP does not mention anything (as far as I can see) that `autocomplete="off"` fails to do so; it seems to be about preventing Chrome from autofilling form fields (such as saved usernames and passwords).

Please revert commit 6611606. Edit: I’ve submitted a PR: #16756.

PR #16756, opened 2015-08-06
Revert "autocomplete='off' no longer works in chrome. Instead, use au…

lydell:
…tocomplete='false'"

`autocomplete="off"` works just fine in Chrome, and Firefox, for the purposes of
disabling the native dropdown of previously entered strings (which would cover
awesomplete's dropdown), while `autocomplete="false"` does not.

This reverts commit 6611606f1105eab3529348f935306c2003c2045d.

Conflicts:
    awesomplete.min.js


jeanphix:
+1 for this one, the `autocomplete` attribute value must be `off`.

LeaVerou:
Merged, thanks!

PR #16773, opened 2015-10-22
Removed quotes around author credit in minified file

jameshibbard:


pazams:
@jameshibbard the minified file is generate by "Makefile" located at project directory. I haven't inspected the reason behind the PR, but in any case the edit should be done at "Makefile".

jameshibbard:
I'm not sure it's anything to do with the Makefile — that seems to be correct.
The problem is that there are two quotes in `awesomeplete.min.js` around the author credit.
If you clone the repo and run make then these disappear.  However, their presence breaks the library for people including it from GitHub directly (e.g. via rawgit). Wasn't sure of the best way to bring this to someone's attention.

pazams:
I verified that when _**I run**_ the make file, the quotes get removed. you are correct on that.
By inspecting the logs, these quotes were added in commit 6611606 by @marc-costello   .
Since that commit had the same version of the make file made by @vlazar , it's fair to assume that the make file behaves differently with these quotes on different systems.

This PR is a good hotfix, but this issue may reappear if isn't treated at the makefile.
@vlazar , please advise.

vlazar:
@pazams I think you are right about `Makefile` being the issue when it was used by @marc-costello to minify. Removing double quotes from banner in https://github.com/LeaVerou/awesomplete/blob/gh-pages/Makefile#L3 might resolve the issue. Unless spaces won't add another one :)

@marc-costello Can you please check, if suggested change fixes the issue on your system?

@LeaVerou Might be a good time to replace Makefile with node (npm scripts, gulp, ...), since it needs to be installed to run the tests anyway. Also, the development process will be simpler and more cross platform.

The Makefile was introduced before npm and karma test runner were merged. Time to retire Makefile?

LeaVerou:
Merged as it fixes the issue for now.
I’m all in for retiring makefile, as long as someone sends a PR. Ideally whatever replaces it should not require too much setup (node is ok as a dependency, no other dependencies plz), to facilitate contributions.

vlazar:
I can do it, since I created Makefile.
The only required setup for contributors in any case would be running `npm install` command.

PR #16776, opened 2015-10-29
expand test-suite

herregroen:
Adds expanded tests for basic functionality. May expand it further when I have time.

Used a separate shared file for repeating tests, seemed the cleanest way to do this but I'm open for suggestions otherwise.


LeaVerou:
Awesome, thanks!

vlazar:
@herregroen You might be interested in #16785

PR #16781, opened 2015-11-23
Allow user to specify pre-existing dropdown element via options param

fractaledmind:
First, let me say that this package is truly awesome! I've been working on some custom autocompleters, comboboxes, and multi-input autocompleters for a work project, and the jQuery UI Autocomplete widget was quickly becoming far more of a hassle than it was worth. Awesomplete made my job _so much easier_. For that I am truly thankful.

In working on my particular use cases, however, I did find the need to add a couple of features. I thought some might be more generally useful, so I'm opening PRs for each.

This PR arose out of my need to create a multi-input autocompleter. Basically, I have four inputs which correspond to 4 sections of an account number. While the ability to override the `filter`, `item` and `replace` methods made the "hard" part quite doable, I found that I had 4 `ul`s when I really only needed the 1. Thus, I added the ability to pass in a `dropdown` option on initialization.

In short, `this.ul` will either be set to the passed in node element or a new `ul` will be created. Regardless, the `ul` dropdown will have the `awesomplete-dropdown` class, which is what the CSS now uses to handle styling.

In order to ensure the dropdown was placed _within_ the container (my own fork incorporates #16743, so the container could be created or pre-existing), I split out the attribute setting logic currently found in `$.create` into a new helper, `$.update`. This allows me to ensure that the hidden attribute is set, the `awesomplete-dropdown` class is set, and the element lives inside the container, even when the element is being passed in.

If there are any stylistic issues or you have any API suggestions, I'm happy to alter this. Or, if you feel this is unnecessary for the base project, I can keep it in my fork happily.

Thanks once again for this,
stephen

### DIFF
--- a/awesomplete.css
+++ b/awesomplete.css
@@ -5,16 +5,16 @@
 	clip: rect(0, 0, 0, 0);
 }
 
-div.awesomplete {
+.awesomplete-wrapper {
 	display: inline-block;
 	position: relative;
 }
 
-div.awesomplete > input {
+.awesomplete-wrapper > input {
 	display: block;
 }
 
-div.awesomplete > ul {
+.awesomplete-dropdown {
 	position: absolute;
 	left: 0;
 	z-index: 1;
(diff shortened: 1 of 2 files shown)


LeaVerou:
Hi there,

Glad Awesomplete made your life a bit easier and thanks for contributing!
I have no problem merging this change (in fact, I usually also have the setting logic separate in my DOM helpers — I call it `$.set()` instead of `$.update()`, so I’d rather keep that name).

However:
- It looks like your editor is a bit opinionated and it messed up the CSS formatting.
- The indentation has been switched to spaces (which I hate) in the part you changed.
- Changing the CSS classes applied is backwards incompatible. People who have already written CSS to customize Awesomplete will see their styling break. I’d rather avoid that, especially for such a small change.

I will close this PR for housekeeping purposes, feel free to reopen if you want to fix the above issues or send a new one :)

PR #16782, opened 2015-11-23
Ensure highlighting matches the filter method chosen

fractaledmind:
This is the second of my PRs. My autocompleters needed to use the `Awesomplete.FILTER_STARTSWITH` method for filtering. However, when I set that as the `filter` method, I found that the highlighting still highlighted matches anywhere in the string. In short, the default `item` function didn't respect the `filter` method.

I solved this problem by creating a `_.RE` variable which is set inside of the two `filter` methods and is re-used inside the default `item` method.

I do see one setback to this approach, however. In order for a user to override either the `filter` method or the `item` method without overriding the other and retain matching behavior, they will need to know about and use this `Awesomplete.RE` lambda. This could be solved simply via documentation, but I'm not certain how flexible this approach is, or whether you would like this. Thus why I don't currently have updated docs in the PR.

I believe that we need a simple way to ensure that the `filter` method and the `item` method are working in parallel, but I'm not yet convinced this is the best approach. So, consider this PR an initial attempt and a space to discuss other/better approaches.

stephen

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -25,10 +25,9 @@ var _ = function (input, o) {
 		filter: _.FILTER_CONTAINS,
 		sort: _.SORT_BYLENGTH,
 		item: function (text, input) {
-			var html = input === '' ? text : text.replace(_.RE(input), "<mark>$&</mark>");
-			return $.create('li', {
-				innerHTML: html,
-				'aria-selected': 'false'
+			return $.create("li", {
+				innerHTML: text.replace(_.RE(input), "<mark>$&</mark>");,
+				"aria-selected": "false"
 			});
 		},
 		replace: function (text) {
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -25,9 +25,10 @@ var _ = function (input, o) {
 		filter: _.FILTER_CONTAINS,
 		sort: _.SORT_BYLENGTH,
 		item: function (text, input) {
-			return $.create("li", {
-				innerHTML: text.replace(RegExp($.regExpEscape(input.trim()), "gi"), "<mark>$&</mark>"),
-				"aria-selected": "false"
+			var html = input === '' ? text : text.replace(_.RE(input), "<mark>$&</mark>");
+			return $.create('li', {
+				innerHTML: html,
+				'aria-selected': 'false'
 			});
 		},
 		replace: function (text) {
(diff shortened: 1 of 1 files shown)


fractaledmind:
Of course, I guess the simplest option would be to add something like a `regexp` or `matcher` option to the `o` hash which both `item` and `filter` look to. Making it explicit in the options hash would make it simpler to document how best to override those methods.

In general, I think it would make `Awesomplete` more flexible if matching, highlighting, and filtering were decoupled and properly layered. Matching is common to both highlighting and filtering, so it ought to be set once and then simply used in those two areas. My own 2 cents are that `filter` should be replaced with `match` or `regexp` (whatever would be best to call the option), which is simply a function (the basic examples would be a `RegExp` lambda, but could also be a function that returns a `Boolean`, e.g. 

```
match.contains   = function(q) { return RegExp($.regExpEscape(q.trim()), 'i'); };
match.startsWith = function(q) { return RegExp('^' + $.regExpEscape(q.trim()), 'i'); };
match.endsWith   = function(q) { return RegExp($.regExpEscape(q.trim()) + '$', 'i'); };
match.equals     = function(text, input) { return text === input; };
```

). Then, highlighting is also private and is called before `item` no matter what, using the set `match` function. Filtering then is also private and always simply calls the match method, whether a `RegExp` object is returned (thus calling `.test()`) or a `Boolean` (thus just passing the Boolean).

The two API options are then `match`/`regexp` and `item`, but `item` only handles the actual DOM element creation, not the `innerHTML` text setting, which uses `highlight` internally to do that.

That's a fairly large scale change, however; so I'm not strongly advocating for that. Simply trying to articulate my thoughts and position on this tricky issue.

LeaVerou:
Hi there,

Thanks for contributing!
Agreed this is confusing, though the proposed solution is a bit awkward. 
Question: If you’re only matching at the start, why do you even need the highlighting? Why not just add some CSS to set the background of `<mark>` to `none`?

fractaledmind:
That is a simple and reasonable solution to my particular issue; however, the larger architectural issue does remain. In short, you cannot override only `item` or `filter` and have sensible behavior. You either have to override neither or both. The conceptual solution is to have a `match` method that underlies both. However one determines if `input` matches `text`, that logic should be used in filtering and in item rendering.

LeaVerou:
Agreed on your assessment of the problem. However, this solution is having multiple Awesomplete instances modify a single static `Awesomplete.RE` variable. I’m not sure how this is either conceptually elegant or simple & reasonable...

PR #16784, opened 2015-11-23
Add combobox advanced example

fractaledmind:
I needed to make a "combobox" and found that adding a `toggle` API method helped with that immensely. Thought it might be more generally helpful.

It handles 3 states:
1. Dropdown is empty
2. Dropdown is hidden
3. Dropdown is open

For 2 and 3, `toggle` simply calls the corresponding `open` or `close` method (ensuring the proper event is fired). For 1, `minChars` is set to `0` (to ensure the full data list is shown) and `evaluate` is called (which also fires the `awesomplete-open` event).


LeaVerou:
Hi Stephen, thanks for contributing!
This looks like it could be a separate plugin or example, since it’s not used internally, it just extends the API. Please make an advanced example with this functionality? That’d be great!
If many people find it useful, we can add it to the API later.

fractaledmind:
I'll update this PR to be a documentation change instead.

LeaVerou:
Thanks! Let me know when you do so I can merge this in :)

fractaledmind:
Sorry for the crazy delay, but life got crazy. I've updated this PR to include only an additional "Advanced Examples" section along with a small number of CSS additions to style to combobox input.

LeaVerou:
Better late than never :) 
Could you please scope the CSS to `#combobox` (or `[id="combobox"]` to avoid the extra specificity) to avoid future conflicts and confusion?

fractaledmind:
Done :)

LeaVerou:
Merged, thanks!

PR #16791, opened 2015-11-29
Don't select items on mouse right-click

TxHawks:
I think it is a bit counter intuitive that right-clicking an item fires the select function.

I suggest only firing the select function if the `mousedown` event is a right click (`evt.which === 1`).


awesomplete.js
```diff
@@ -95,7 +95,7 @@ var _ = function (input, o) {
 				li = li.parentNode;
 			}
 
-			if (li) {
+			if (li && evt.which === 1) {  // Only select on left click
```
LeaVerou: Given that `evt.which` is non-standard, I'd rather do `evt.which !== 2 && evt.which !== 3` or `[2,3].indexOf(evt.which) === -1` or something like that.

LeaVerou:
Thanks! See comment.

awesomplete.js
vlazar: Or https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button ?

TxHawks: My bad. 

I guess we should probably go with the standard and less convoluted `evt.button === 0` which [seem to be well supported](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button).

LeaVerou:
Yes! Nice find!
Merged, thanks!

PR #16792, opened 2015-11-29
Code Climate

vlazar:
[Code Climate](https://codeclimate.com/) analyzes code with tools like `eslint`, `csslint`, and others to keep an eye on code style and quality. It shows code coverage reports, so if code coverage degrades or improves, you'll notice. It can be integrated with GitHub to report this information back.

Let's start with the following features added by this PR:
- turn on `eslint` (with recommended config + smart-tabs) and `fixme` reporters
- generate code coverage when running tests
- send code coverage results to Code Climate on each Travis CI build
- Code Climate badges in README

Before merging this PR:
1. Create an account on https://codeclimate.com/
2. Add `LeaVerou/awesomplete` GitHub Repository to Code Climate
   https://codeclimate.com/github/signup
3. Copy `CODECLIMATE_REPO_TOKEN` value from this page on Code Climate:
   https://codeclimate.com/repos/YOUR_ADDED_REPO_UNIQUE_ID/coverage_setup
4. Add environment variable `CODECLIMATE_REPO_TOKEN` with value copied on step 2 on this page on Travis CI:
   https://travis-ci.org/LeaVerou/awesomplete/settings

Then merge this PR.


LeaVerou:
Hey, thanks for the contribution!

It's reporting an error about mixed tabs and spaces because it's buggy when tabs are used for indentation and spaces for alignment, which is my preferred style. See: http://lea.verou.me/2012/01/why-tabs-are-clearly-superior/

In general, linting is nice as an aid, but not when it becomes too opinionated and doesn't let you bypass its recommendations. I’m a bit unclear on how exactly this system handles this.

vlazar:
> In general, linting is nice as an aid, but not when it becomes too opinionated and doesn't let you bypass its recommendations.

Don't worry, **you can bypass recommendations**, and/or you can change any `eslint` setting in https://github.com/LeaVerou/awesomplete/pull/16792/files#diff-1dc6ee56b778cd91e0327b52aaeaa1b9R1 It's a config for Code Climate, not for Travis CI,  but can be used locally with `eslint` too. All these checks can be automated to be run by contributors with 1 command BTW. So they can know something is not right even before submitting a PR (if they read a contributor guide of course).

**Travis CI build won't fail because of this**, but you will see the info provided by Code Climate on Code Climate. Wouldn't it be nice, if it checks for spaces vs tabs for you, so you wouldn't have to?

And you can configure the list of check engines in https://github.com/LeaVerou/awesomplete/pull/16792/files#diff-0655d430acfa953cae08b6619689d18cR1 There is a `csslint` support on Code Climate by the way, which can be added and configured similar to `eslint`.

My original intention with this PR was to provide code coverage info. It good to know, if a new feature comes not covered by tests.

But Code Climate does other good stuff too, so why don't have it for free? It should provide info for send PRs too, I believe, so you'll see immediately, if someone uses spaces for indentation ;)

vlazar:
Oh, maybe I wasn't clear - `eslint` gives no errors or warnings with current config :)

LeaVerou:
Ok, I’m on step 3, where do I find the repo's unique id in Code Climate?

vlazar:
To copy CODECLIMATE_REPO_TOKEN
1. Hover over your repo and click Settings

1. Then click Test Coverage on the left
1. Then click JavaScript tab

.eslintrc
```diff
@@ -0,0 +1,253 @@
...
+  no-label-var: 0
+  no-labels: 0
+  no-lone-blocks: 0
+  no-lonely-if: 0
+  no-loop-func: 0
+  no-mixed-requires:
+  - 0
+  - false
+  no-mixed-spaces-and-tabs:
+  - 2
```
LeaVerou, line 55: What does this value mean?

LeaVerou:
I’m ok to merge (I finished the 4 steps), but not sure why the Travis build is failing. Could you look into it?
Also, see comment.
Thanks!

.eslintrc
vlazar, line 55: http://eslint.org/docs/user-guide/configuring#configuring-rules

This is for all config rules, not only this one:
0 - turn the rule off
1 - turn the rule on as a warning (doesn't affect exit code)
2 - turn the rule on as an error (exit code is 1 when triggered)

Smart tabs meaning below is described here:
http://eslint.org/docs/rules/no-mixed-spaces-and-tabs.html

vlazar:
The gh-pages from origin Travis CI build was failing when it was merged into this PR, so this build just has no recent fixes from origin.

Sorry for confusion, I should've merged latest gh-pages. Will do now.

.eslintrc
LeaVerou, line 55: This will probably trigger pointless errors since I do, in fact, mix tabs and spaces on purpose. I use tabs for indentation and spaces for alignment, a style that surprisingly few tools have caught up to, despite the number of people using it.
So, if that's the case, we should change that to a 0...

LeaVerou:
Merged now, but please address the spaces vs tabs issue? Thanks!!

.eslintrc
vlazar, line 55: I've already added the `- smart-tabs` option below, which fixed the error shown by eslint locally for exactly your case - tabs for indentation and spaces for alignment. I also see no errors from eslint at https://codeclimate.com/github/LeaVerou/awesomplete/issues/categories/bugrisk about mixed spaces and tabs.

vlazar:
Thanks!

I'll do one more PR for CodeClimate as at least newly added `prism/prism.js` should be removed from the list of files CodeClimate runs metrics on. Not sure if we want to keep `index.js` though on this list in addition to `awesomplete.js`. It's kind of not the source of the lib, but we still might want to keep an eye on it and see if there are any errors.

@LeaVerou So what do you think? Should we remove both `prism/prism.js` and `index.js` from the list or just the `prism/prism.js`?

vlazar:
See #16810

.eslintrc
LeaVerou, line 55: Cool, thanks!

Issue #16788, opened 2015-11-25
Handling an array of objects


travpeck:
Would it be feasible to have the feature where you could have an array of objects like such:

```
[{ name="Johnny Appleseed", id="100" }, { name="Paul Bunyan", id="200" }]
```

And be able to show the _name_ property in the list but also associate the _id_ property with it once it is selected? Basically hold a reference to the record of the selected item.

Example:

```
var input = document.getElementById("myinput");
new Awesomplete(input, {
    list: [{ name="Johnny Appleseed", id="100" }, { name="Paul Bunyan", id="200" }],
    display: 'name'
});

input.on('awesomplete-selectcomplete', function() {
    // Ability to get the selected value, such as 'Johnny Appleseed'
    // But also get other record information (such as id)
};
```

vlazar:
**To everyone interested in differentiating value from displayed text, aka key/value feature, aka array of objects feature.**

The code is ready and it's almost 100% backward compatible. To ensure everything working fine for you we need your help with testing it in the wild before it's released. See this PR: [Separate label/value for each suggestion in list](https://github.com/LeaVerou/awesomplete/pull/16852)

vlazar:
Separate `label/value` is landed #16866
It is also possible to initialize list with separate `label/value` via `<datalist>` or `<ul>`.

Read about different `label` and `value` at the end of [Basic usage](http://leaverou.github.io/awesomplete/#basic-usage) section and about new `data` method at the end of [Extend](http://leaverou.github.io/awesomplete/#extensibility) section.

PR #16794, opened 2015-12-05
Pass the original event on select

gabesoft:
Sometimes we need to know the exact element that was clicked on when selecting an item in the dropdown. This PR adds the original event to the select data when firing the  `awesomplete-select` event.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -96,7 +96,7 @@ var _ = function (input, o) {
 			}
 
 			if (li && evt.button === 0) {  // Only select on left click
-				me.select(li);
+				me.select(li, evt);
 			}
 		}
 	}});
@@ -190,7 +190,7 @@ _.prototype = {
 		$.fire(this.input, "awesomplete-highlight");
 	},
 
-	select: function (selected) {
+	select: function (selected, originalEvent) {
 		selected = selected || this.ul.children[this.index];
 
 		if (selected) {
@@ -200,7 +200,8 @@ _.prototype = {
 				text: selected.textContent,
 				preventDefault: function () {
 					prevented = true;
-				}
+				},
+				originalEvent: originalEvent
 			});
 
 			if (!prevented) {


LeaVerou:
I don't see the use case, but it's a pretty small change, so merging. :)
Thanks!

TxHawks:
I think can be very useful, and a better solution to #16788 than sending dataset to the `awesomplete-select` and `awesomplete-selectcomplete` events, as I suggested in #16795. 

However, this introduces a bug, where selecting an item with the keyboard and with a mouse does not have the same results. Right now, the original event is only sent back when selecting an item with a mouse. The problem is, that when selecting an item with the keyboard, the event's target is the input, and therefore not useful.

Since we don't really care about the original event, but only about the target, I suggest assigning the `li` to `evt.originalTarget` and then using that instead of the entire event.

Another issue, is that I think that if we are exposing the selected element to the `awesomplete-select` event, we should also expose it to `awesomplete-selectcomplete`.

I have updated #16795 with these changes, as well as updated documentation.

vlazar:
@gabesoft I know this is already merged in, but I'm trying to combine this change with other related features.

Can you please clarify what did you mean by:

> Sometimes we need to know the exact element that was clicked on when selecting an item in the dropdown.
- did you mean exact element is a `<li>` element being selected when you click dropdown item with mouse?
- do you really care about original event object, or just an HTML element being selected?
- if you do care about original event object, what event do you think should it contain in `awesomplete-select`, when selecting an item with keyboard instead of mouse?

gabesoft:
In my use case I have a delete button next to each item in the autocomplete list. So, I need to know when a user clicks that button rather than the whole item. Here's an example of how I'm using this https://github.com/gabesoft/xandar/blob/master/lib/components/tags-input.jsx#L69

vlazar:
Looks like in example you linked to you only need an `event.target` e.g. HTML element, not the whole event object, right?

**What if instead of original click event object the clicked element would be passed into `awesomplete-select` event? Would this work for you?**

The issue with event object is that for the case when you select an item with keyboard there is no event passed currently. Also passing keydown event in that case sounds not useful as it wouldn't be related to the clicked item at all.

vlazar:
Changing `awesomplete-select` to pass something like `target: clickedElementLikeButton` (with different property name, as target should refer to input really) instead of `originalEvent: originalEvent` would also simplify code for you:

https://github.com/gabesoft/xandar/blob/master/lib/components/tags-input.jsx#L83-L92

Can be replaced with:

``` javascript
      input.addEventListener('awesomplete-select', event => {
        const value = awesomplete.replace(event.text);
        // const target = (event.originalEvent || {}).target || {}; // REMOVED LINE
        if (target.tagName === 'I') {
          actions.deleteTag(value);
        } else {
          input.value = value;
          el.setState({ tag: value });
        }
      });
```

gabesoft:
This will work for my use case. Not sure if we'll need other info from the original event at some point. But, I guess we don't have to worry about that now.

vlazar:
Thanks for answering, this will help a lot to satisfy other requests too.

vlazar:
@gabesoft FYI This is landed in https://github.com/LeaVerou/awesomplete/commit/eb205864f435a3b32499b5291e3685f11a2e786e#diff-0861bfe5a8e20a86e84f647fe9e26358R205

Instead of `originalEvent` you get `origin` element which is `evt.target` from mousedown event, or just selected LI element (when selected with keyboard or via API). So it's guaranteed to be present on event object.

You now can do:

``` javascript
      input.addEventListener('awesomplete-select', event => {
        const value = awesomplete.replace(event.text);
        // const target = (event.originalEvent || {}).target || {};
        // if (target.tagName === 'I') {
        if (event.origin.tagName === 'I') {
          actions.deleteTag(value);
        } else {
          input.value = value;
          el.setState({ tag: value });
        }
      });
```

gabesoft:
Great, that will work. Thanks.

vlazar:
The name of event property is now `originalTarget` instead of `origin` https://github.com/LeaVerou/awesomplete/commit/1f86aaf981910686e11ffea47c24c8838a17d7d8

LeaVerou:
Cool!

PR #16798, opened 2015-12-11
Fix issue when an element with ID of `exports` is defined in the browser.

bnjmnt4n:
Elements with IDs automatically have their IDs exposed. This causes errors where `module` is not defined, and yet `module.exports = _` is run. This also aligns the export with UMD.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -383,7 +383,7 @@ if (typeof self !== "undefined") {
 }
 
 // Expose Awesomplete as a CJS module
-if (typeof exports === "object") {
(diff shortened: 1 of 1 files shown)


bnjmnt4n:
Ping? ;)

LeaVerou:
Merged, thanks!
Though not sure why anyone would have an id of exports. :P

bnjmnt4n:
It’s came up before on some production apps (although it’s true – not very often), and also changing the exports code this way aligns with UMD.

PR #16813, opened 2016-01-08
Fix #16808. Input loses focus if item selected with mouse.

### LINKED ISSUE #16808: Selecting an Autocomplete suggestion should keep the focus on the input box
It would be nice to have Autocomplete to keep focus on the box after selecting the first item.

vlazar:


awesomplete.js
```diff
@@ -96,6 +96,7 @@ var _ = function (input, o) {
 			}
 
 			if (li && evt.button === 0) {  // Only select on left click
+				evt.preventDefault();
```
vlazar, line 99: @LeaVerou I think it's safe to prevent default action for `mousedown` event on LI element here. Maybe you can se any cases where it might hurt?

LeaVerou, line 99: Nope, I can't think of any!

LeaVerou:
Thanks, merged!
And thanks so much in general for all your hard work around here!! I'm so glad I made you a collaborator, you rock!

vlazar:
Thanks Lea, no problem, really like your stuff.

I've been thinking recently about having a way to deal with (small) PRs faster... So I'll post this off-topic comment here.

Do you think something like this might be useful and save you some time dealing with PRs?
- ♥ 2 in menu bar means 2 small PRs (green heart is for Small PRs group)
- ♥ for any PR item in menu means PR is mergeable
- ♡ for any PR item in menu means PR is not mergeable 

prs-1

prs-2

This is basically just a list of PRs sorted by the number of changes. This can be changed to include as much info as necessary, even use multiple lines for each PR. For the concept I wanted it to be very minimal. But, of course, can be as much info as needed, or as customizable as needed.

LeaVerou:
This looks awesome!!! Not sure about the icons (a bit cryptic) but the idea is fantastic! Can it work for multiple projects?

vlazar:
Sure, I can extend it to work for multiple projects too. I've just tried to imagine how it can look like and what info it needs to show.

Glad you like the idea. I think this can be useful for other people too.

I'll create a repo then once I have free time. Let's discuss details, like how simple or how extensible this should be in new repo or whatever place works for you for discussing ideas. Happy to hear your list of features for this tool.

TL;DR I'm terrible with UI/UX like many devs, so I just plan to add themes support. Maybe you'll create an awesome theme then.

LeaVerou:
Good UI/UX is not about skinning though. This is a misconception that many devs have. Good design is not how it looks, it's how it works, primarily.
Also, I never understood why devs who are "terrible with UI/UX" don't try to study the stuff. There are tons of interesting books and they really helps. It's not something that you either have or you don't, knowledge goes a long way. 
Anyhow, I'd be happy to give more detailed feedback once there's a repo. I think it's a great idea that would help tons of project maintainers, especially if done well.

vlazar:
Used the wrong word. Themes, but not for colors only... Also what info about each repo/group/etc. and in what way to show, how to sort, etc. For quicker results I can go with your use case only, or start thinking about generic case from the beginning.

vlazar:
Ok, let's start a discussion here https://github.com/vlazar/a-priori-tizer/issues/1

I recognize it can be made generic enough to prioritize any list of tasks later by some criteria, like GitHub issues, hence the generic name for this project. Please, share your feedback.

For now lets concentrate on your personal minimal feature set for GitHub PRs only.

PR #16816, opened 2016-01-09
Pass selected node to event

ghost:
Passing just text is good but passing the entire element allow more flexibility, i.e. one can know on what link user clicked

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -197,6 +197,7 @@ _.prototype = {
 			var prevented;
 
 			$.fire(this.input, "awesomplete-select", {
+				selected: selected,
 				text: selected.textContent,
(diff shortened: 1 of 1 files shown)


TxHawks:
@sleewoo -
The selected node is already passed as of #16794, sort of. The original event is passed through the originalEvent property ([see here](https://github.com/LeaVerou/awesomplete/blob/gh-pages/awesomplete.js#L204)), which allows you to look at originalEvent.target

However, there is a bug, because it is only passed when an item is selected using a mouse, not the keyboard. #16795 fixes this, but have  not been accepted yet.

ghost:
thanks, somehow missed `originalEvent`,
hoping your fix will be accepted soon.

Issue #16818, opened 2016-01-15
Redo recently merged feature


vlazar:
@LeaVerou I'm starting work on key/value (aka array of objects) feature related PRs and issues and badly need you opinion on this.

Recently merged PR https://github.com/LeaVerou/awesomplete/pull/16794/files added `oginalEvent` to `awesomplete-select` event. However, the are several issues with it:

``` diff
...
-   select: function (selected) {
+   select: function (selected, originalEvent) {
...
    $.fire(this.input, "awesomplete-select", {
-                               }
+                               },
+                               originalEvent: originalEvent
```
1. It changes the public API
2. It only passes `oginalEvent` on `mousedown` event, but not on `keydown` (bug)
3. It passes the whole event object to `awesomplete-select` event for the original use case

> to know exact element that was clicked on when selecting an item in the dropdown

I think that:
1. Should be reverted, so we pass data via instance property in order not to change public API
2. Should be fixed and the same data should be passed to `awesomplete-select` caused by keypress
3. Only pass `event.target` instead on `event`, unless you see use case to pass the whole `mousedown` or `keydown` event. [I've discussed with PR author, and `event.target` is enough for his use case](https://github.com/LeaVerou/awesomplete/pull/16794#issuecomment-171731830)

LeaVerou:
Agree with 2 & 3. :+1: 
Changing the public API is ok, as long as it's backwards compatible (and this is). But of course, it should be done carefully.

vlazar:
Are you absolutely sure about 1? The `originalEvent` argument seems to have no meaning in the context of public API:

`completer.select(selected, originalEvent)`

It looks like it was only added because it was easier to pass an argument, and maybe author didn't realized that it also changes public API and `originalEvent` in the context of public API is meaningless.

LeaVerou:
I wasn't saying that passing `originalEvent` around was a good idea. Just that (generally speaking) API changes are not the devil's spawn, as long as they're backwards compatible.
In this case I agree that it was a clumsy solution.

vlazar:
Addressed 2 & 3 in #16819 (please do not merge, see questions in it)

PR #16819, opened 2016-01-16
Feature/pass target to awesomplete select

vlazar:
Addresses #16818

I'm not happy with option named `target` though.

Since an `awesomplete-select` is fired on INPUT, inside it `this`, `event.target` and `event.currentTarget`  are references to INPUT originally. In click `target` is a reference to clicked element, but in `awesomplete-select` it may be misleading.

Eventually we need to pass 2 elements to `awesomplete-select` event for new features:
- first one is in this request - exact clicked item (child of LI or LI itself)
- second is always LI (selected element)

I can't come up with good aligned names for those 2 new fields in `awesomplete-select` event.

Any ideas?

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -190,7 +190,7 @@ _.prototype = {
 		$.fire(this.input, "awesomplete-highlight");
 	},
 
-	select: function (selected, target) {
+	select: function (selected, origin) {
 		selected = selected || this.ul.children[this.index];
 
 		if (selected) {
@@ -201,7 +201,7 @@ _.prototype = {
 				preventDefault: function () {
 					prevented = true;
 				},
-				target: target || selected
+				origin: origin || selected
 			});
 
 			if (!prevented) {
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -87,7 +87,8 @@ var _ = function (input, o) {
 	$.bind(this.input.form, {"submit": this.close.bind(this)});
 
 	$.bind(this.ul, {"mousedown": function(evt) {
-		var li = target = evt.target;
+		var target = evt.target;
+		var li = target;
 
 		if (li !== this) {
 
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -87,7 +87,7 @@ var _ = function (input, o) {
 	$.bind(this.input.form, {"submit": this.close.bind(this)});
 
 	$.bind(this.ul, {"mousedown": function(evt) {
-		var li = evt.target;
+		var li = target = evt.target;
 
 		if (li !== this) {
 
@@ -96,7 +96,7 @@ var _ = function (input, o) {
 			}
 
 			if (li && evt.button === 0) {  // Only select on left click
-				me.select(li, evt);
+				me.select(li, target);
 			}
 		}
 	}});
@@ -190,7 +190,7 @@ _.prototype = {
 		$.fire(this.input, "awesomplete-highlight");
 	},
 
-	select: function (selected, originalEvent) {
+	select: function (selected, target) {
 		selected = selected || this.ul.children[this.index];
 
 		if (selected) {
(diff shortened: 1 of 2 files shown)


TxHawks:
I think that 
1. both `target` and `selected` should be passed to `awesomplete-select`, though I guess it's not a must, since you could always get to it using `target.parentNode()` or `target.closest()`.
2. `awesomplete-selectcomplete` should also be privy to `target` (and possibly `selected`, if it's added so `awesomplete-select`

vlazar:
@TxHawks
1. I've been thinking about using `target` and `selected` names, except that `target` is INPUT originally. Personally I'm leaning towards passing both, since `selected` is essential and `target` is nice to have for advanced use. Also IE/Edge currently is lacking `target.closest()` and until awesomplete goes with polyfills it may be not an option to keep this for users to deal with.
2. Seems your text is cut. I agree that other event's options should be reviewed, and we need to pass new options where it makes sense).

With this PR and your #16795 landing it will be easier to do a frequently requested key/value (aka arrays of objects) feature among other things.

TxHawks:
@vlazar

It wasn't cut, just had a typo. The last `so` should have been `to`.

Tjanky you for doing all this work, by the way

LeaVerou:
Maybe an `item` property for the LI? Not sure, need to think about this more.

Btw, minor code nitpick: Please don't use variables when they are only used once. It's an indirection for humans reading the code (as they need to pointlessly look up the variable definition) and bloats the code size too. In this case, using `evt.target` in both lines would have been fine, no need for a `target` variable just to save one property reference. I know it's supposed to be marginally faster, but it's premature optimization and readability matters more.

vlazar:
> Maybe an item property for the LI? Not sure, need to think about this more.

I've been thinking about using `item` later as a good property name for item data itself, when items are objects instead of strings. We have `text` property for item data now, but this name only makes sense when each item is a string. Would be nice change or addition in my opinion.

> In this case, using evt.target in both lines would have been fine, no need for a target variable just to save one property reference.

I'll change this. Wasn't intended as a way to save a property reference though. The `target` and `li` are conceptually different. The `li` just starts from `target` originally and became guaranteed `li` in a while loop later going up the DOM tree. We might want to replace this with `element.closest()` polypill later if polyfills are an option as in Bliss.

vlazar:
How about an `origin` property instead of `target` for exact clicked element (child of LI or LI itself)? It's like `target` element in `mousedown` event becomes `origin` element of `awesomplete-select` event.
cc @TxHawks

vlazar:
Maybe `origin` for exact clicked element and `element` for selected LI element?

The more I think about it, `item` property with item data seems like the most useful property to add. I guess my current vision of dream API:

``` javascript
// existing option; I prefer to replace/deprecate it in 2.0 with next option
text: itemText,
// most useful property (text when items are strings, object if items are array of objects)
item: itemTextOrItemObject,
// allows to do something with selected element
// less useful, especially if item property above added
element: selectedItemElement,
// very specific use case
// the least useful, if at all
// maybe there is a better way for original use case for added `originalEvent` option
origin: clickedChildOfItemElement
```

vlazar:
@LeaVerou Can you please look at PR in the current state, and if you can't think of better name I'll clean this up and merge?

And my comment above https://github.com/LeaVerou/awesomplete/pull/16819#issuecomment-173197709 with possible properties for next PRs.

I think `item` would be great addition. For 2.0 it can be even a replacement for `text`. We can pass it everywhere we use `text` property or argument. It aligns well with having a list items in different formats (e.g. objects in addition to strings we have now).

LeaVerou:
I'm a bit confused with all the PRs, but I just left a comment in the other one.

vlazar:
Sorry for confusion. Just let me know if this PR should be merged. This PR is only about 2 & 3 as you already agreed in https://github.com/LeaVerou/awesomplete/issues/16818#issuecomment-172209898

LeaVerou:
Sure, this is fine.

vlazar:
Closing in favor of #16823

Issue #16821, opened 2016-01-21
API consistency. Use item instead of text everywhere


vlazar:
Continues #16818

With list of string items everything works out of the box right now.

There is also a frequently requested feature to have items as objects (key/value pairs). Making it work on top of current API is problematic.

Currently we pass list item (either a string or an object) to:

``` javascript
// text is actually a list item, even if it's an object
filter: function (text, input)
item: function (text, input)
// a and b both items (text if list is array of strings, object if it's array of objects)
sort: function (a, b)
```

And here we have text (so items as object won't work):

``` javascript
// text here always text as it's called as this.replace(selected.textContent);
replace: function (text)
// and we also have text property in here
$.fire(this.input, "awesomplete-select", {
    text: selected.textContent,
```

I think this needs to be changed so that when we have list of objects both `replace` and `awesomplete-select` would also get the whole item data instead of text only. This change would make features like key/value support much easier to implement.

If we go with this, then the question is how we implement this API change.

We can make it as non breaking change in `1.x` (possibly showing deprecation warning in console too):

``` javascript
replace: function (text, item)
$.fire(this.input, "awesomplete-select", {
    text: selected.textContent,
    item: this._list[this.index]
```

And cleanup API in 2.0:

``` javascript
replace: function (item)
$.fire(this.input, "awesomplete-select", {
    item: this._list[this.index]
```

However in this case `replace` signature changes twice. So a better option would be to introduce new `item` instance property, which would be a selected item data (text or object) - essentially `this._list[this.index]`. We can even have `replace: function ()` without arguments in 2.0 since `this.item` would be accessible on any instance. So 2.0 API can be:

``` javascript
// just use this.item the same way this.input is used currently
replace: function ()
// still need to pass it to event handler
$.fire(this.input, "awesomplete-select", {
    item: this.item
```

See also related discussion #16819

TxHawks:
+1

LeaVerou:
Idea: How about we pass an object that has properties with all the stuff we need, and a `toString` method that returns the string the 1.x API returned? That way it will be a non-breaking change for most use cases as it will return what's expected any time it's converted to a string for any purpose.

vlazar:
Do you mean wrap `item` so that it'll have `toString` method and pass it into `replace`?

``` javascript
replace: function (item) {
    this.input.value = item.toString();
}
```

Need to think more on this. It might work for `replace`.

[...]

PR #16829, opened 2016-01-28
Pass instance to configure instead of call(this)

vlazar:
Why use `configure.call(this, defaults, o)` if we could just do `configure(this, defaults, o)`?

This also reduces minified code size.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -18,7 +18,7 @@ var _ = function (input, o) {
 
 	o = o || {};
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
I guess that makes sense. Thanks!

vlazar:
Thanks!
Any chance you can look at https://github.com/LeaVerou/awesomplete/issues/16821#issuecomment-174310187 so that I can continue working on the key/value related stuff?

PR #16830, opened 2016-01-29
Expose item & replace as static properties.

vlazar:
Makes it possible to configure these properties globally.
Simplifies tests. We can test static properties in isolation.


LeaVerou:
I don't see that many use cases, but sure, why not. It's a small change and it might help someone.
Thanks!

vlazar:
Cool. Tests will come soon then.

vlazar:
Added tests in #16834

PR #16833, opened 2016-01-29
added option to complete by pressing tab

Nigelli:
This allows the user to confirm selection when tabbing away from the input box instead of first having to press enter.


LeaVerou:
Thanks! However, we need to also check that `autoFirst` is not on.

silverwind:
I'd recommend also checking for the shift key, so moving the focus through shift+tab is still possible. I use `if ((c === 13 || (c === 9 && !evt.shiftKey)) && me.selected)` right now.

LeaVerou:
Closing in favor of #16879.

PR #16834, opened 2016-01-30
Tests for Awesomplete.ITEM and Awesomplete.REPLACE

vlazar:


LeaVerou:
Thanks! Btw you don't need to send PRs and merge them yourself for things like that, just commit directly :)

vlazar:
Sure, but doing PR also goes via TravisCI and Code Climate. Might also go via Sauce Labs cross browser testing one day. So it's safer :)

LeaVerou:
Ah, I see. Ok then! :)

vlazar:
Just realized, maybe you wanted less GitHub notifications noise. I'll keep this in mind then.

PR #16835, opened 2016-01-30
Add item data to awesomplete-select event

vlazar:
I made this PR as minimal as possible just to add new property to `awesomplete-select` event.

However I think if this one is merged, later we should expose `this._suggestions` as public `this.suggestions` property. It's a list of current matched items data. With just this 1 additional property we can do the following in instance methods:

``` javascript
// get current selected item data
this.suggestions[this.index];
// the same way as we can do now to get current selected item DOM element
this.ul.children[this.index];
```

Of course we can go further and add public instance properties to match suggested event properties in your comment https://github.com/LeaVerou/awesomplete/issues/16821#issuecomment-176906184
This will simplify code and be more convenient public API.

``` javascript
// same as above
this.data;
this.element;

// or since we already have this.selected:
this.selectedData;
this.selectedElement;
```

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -189,7 +189,7 @@ _.prototype = {
 		if (selected) {
 			var allowed = $.fire(this.input, "awesomplete-select", {
 				text: selected.textContent,
-				data: this._suggestions[elementSiblingIndex(selected)],
+				data: this.suggestions[$.siblingIndex(selected)],
 				origin: origin || selected
 			});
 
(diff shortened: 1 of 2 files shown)


vlazar:
@LeaVerou I believe now I have a very good implementation for data support in different formats (key/value objects, arrays, etc.). But I need `this.data` or `this.suggestions` or at least `this._suggestions` from here to move further with small simple PRs.

TxHawks:
O'm also waiting for this to update #16795 accordingly

vlazar:
@TxHawks Please don't update your #16795 yet when (and if) this PR is merged. The next PR I have in mind would make the whole key/value story much simpler.

If all goes as I imaging it, we can split your PR into 2:

The one that adds selected element property to `awesomplete-select` event. It looks like Lea is in favor of `element` as property name for that one. I guess we should skip `awesomplete-selectcomplete` event for now to finalize all property names first.

The second one with example of how to work with array of objects. This one can be totally unrelated to added event properties if my second PR will be accepted. Or maybe it can show the usage of event properties in addition to just using an array of objects.

Smaller PRs have better chance to be merged soon and this will cleanup the commit history of #16795 since it looks like the final implementation will look very different from what it was in the beginning.

TxHawks:
@vlazar :thumbsup:

LeaVerou:
Ok with merging once the helper function is added on `$` (and renamed to `siblingIndex`).

PR #16860, opened 2016-02-26
Matching Diacritics

joshuabyler:
Ran into an issue where if a charcter had diacritics (accents), then it does not match.  Added a function that matches characters with diacritics to the matching character without it.  Also wrote a test to make sure that it is working.


vlazar:
Thanks for contribution! It's super cool you've added tests for new feature.

Some notes from quick glimpse into the code:
- you've used `text.latinise()` in `FILTER_CONTAINS` and `FILTER_STARTSWITH`, but not in `SORT_BYLENGTH`
- you've only replaced diacritics in item's text, but not input value; what if user inputs text with diacritics?
- the `String.prototype` should not be modified, `latinise()` can be a local function
- there are a lot of unrelated formatting changes, which makes this PR harder to read, please keep existing formatting style

I'm not sure, if this approach would work for every language. E.g will sorting work correctly in any language? Different languages have different rules. Internationalization is hard.

@LeaVerou Do you think diacritics should be handled in Awesomplete itself by default?

We might want to look into https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl in the near feature, as browser support is not quite there yet.

joshuabyler:
I'm really sorry about all the stupid formatting changes, my ide just changed all of that crap on me.

You make a great point about the input needing to be cleaned up too, I could add that, and I could add the function to the sort by length function.  I'll also remove latinise() from the String prototype.  Should have those changes soon.

Also, about people entering different languages, you're right about i18n being really hard.  It's tough figuring out all the different possible things that could be entered, and then making sure they match.  I was just trying to address diacritics, but if there is an option that would fix it for different languages too that would be awesome

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 coverage
+.idea
```
vlazar, line 3: This can be in a local git ignore, since it's not for all users.

vlazar:
Don't put to much effort into this yet. Let's discuss first.

I think you can clean up formatting and remove latinise() from the String prototype and wait for Lea's comments on this, since she is the project owner. I can't decide by myself if this will be merged, or not.

joshuabyler:
cool thanks for the quick response

vlazar:
> Also, about people entering different languages, you're right about i18n being really hard. It's tough figuring out all the different possible things that could be entered, and then making sure they match.

And we probably should not do it. For example [this Intl polyfill is 30k minified](https://github.com/andyearnshaw/Intl.js)

LeaVerou:
Sorry to be a spoilsport, but I think this should be a plugin.
(no strong opinion though, I'm open to being convinced otherwise)
Thanks for the work @heronblade!
We should have a plugin directory, as the "advanced examples" list is getting too long.

joshuabyler:
hey no problem, i completely understand.

I'm still going to clean it up, and try and make it better.  By the way @LeaVerou awesomplete is pretty freaking cool, it has been really nice working with it, and in fact the company I work for just released it on Steinway's new Artist [page](http://steinway.com/music-and-artists)

LeaVerou:
Thank you @heronblade, glad you like it! And definitely do work on it, this is a common request, so we'll use your code one way or another. Even if it doesn't make it to the core library, it will definitely be added as a plugin, once we have a plugin directory.

> I can't decide by myself if this will be merged, or not.

Just noticed this. For small, noncontroversial PRs, you can (this one is neither though :)

.gitignore
eliasgonzalezugalde: How can I enable this feature? I need that awesomplete ignore ñ - á - é - í - ó - ú, etc

PR #16866, opened 2016-03-12
Separate label/value

vlazar:
This is a cleaned up version of #16852 (sequence #16821 -> #16851 -> #16852 -> this PR)

It is now possible to show a suggestion label in completer, but insert a suggestion value into the input instead.

In addition to String as before, each list item now can also be:
- a `{ label, value }` Object
- a `[ label, value ]` Array

To show full country name in completer, but insert country code into the input you can use these items:
- `{ label: "United States", value: "US" }`
- `[ "United States", "US" ]`

Despite this data format change, old code will continue to work as before. This is taken care by `Suggestion()`. It uses `label` property automatically when string is expected anywhere in the API.

One known issue is that accessing suggestion's characters by index won't work with old API. It's easy to fix though. Instead of `item[idx]` use `item.value[idx]` or `item.label[idx]` directly.

In addition to default support for String/Object/Array items, we also add `data` method, which can be used to support any additional custom item formats and to generate data dynamically, as in changed Email example. The only thing you need to do in this case is to return item in any of String/Array/Object formats supported by default.

Initialization via `<datalist>` or `<ul>` with separate `label/value`  is supported as well.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -118,9 +118,18 @@ _.prototype = {
 			list = $(list);
 
 			if (list && list.children) {
-				this._list = slice.apply(list.children).map(function (el) {
-					return el.textContent.trim();
+				var items = [];
+				slice.apply(list.children).forEach(function (el) {
+					if (!el.disabled) {
+						var text = el.textContent.trim();
+						var value = el.value || text;
+						var label = el.label || text;
+						if (value !== "") {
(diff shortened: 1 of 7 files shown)


LeaVerou:
Thanks for your continued efforts @vlazar! Nice work!

PR #16867, opened 2016-03-12
Pass suggestion item to awesomplete-selectcomplete and awesomplete-highlight events

vlazar:
Similar to `awesomplete-select` event we need to know suggestion item in `awesomplete-selectcomplete` too.

I'm not sure if we also need `origin: origin || selected` same as in `awesomplete-select` event.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -190,7 +190,9 @@ _.prototype = {
 			this.status.textContent = lis[i].textContent;
 		}
 
-		$.fire(this.input, "awesomplete-highlight");
+		$.fire(this.input, "awesomplete-highlight", {
+			text: this.suggestions[this.index]
+		});
 	},
 
 	select: function (selected, origin) {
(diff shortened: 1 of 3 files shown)


awesomplete.js
```diff
@@ -201,15 +201,19 @@ _.prototype = {
...
 				origin: origin || selected
 			});
 
 			if (allowed) {
-				this.replace(this.suggestions[this.index]);
+				this.replace(suggestion);
 				this.close();
-				$.fire(this.input, "awesomplete-selectcomplete");
+				$.fire(this.input, "awesomplete-selectcomplete", {
+					text: suggestion
```
vlazar, line 217: Should we add `origin: origin || selected` too as in `awesomplete-select` above?

awesomplete.js
```diff
@@ -201,15 +201,19 @@ _.prototype = {
 		}
 
 		if (selected) {
+			var suggestion = this.suggestions[this.index];
+
 			var allowed = $.fire(this.input, "awesomplete-select", {
-				text: this.suggestions[this.index],
+				text: suggestion,
 				origin: origin || selected
```
vlazar, line 210: Should we add same property to `awesomplete-selectcomplete`?

vlazar:
And we also have `awesomplete-highlight` event.

I think it makes sense to add `text` property to this event as well?

LeaVerou:
Hi @vlazar, this seems like a pretty minor change, so you could merge it yourself if you wanted (of course it's fine to also post it for feedback! :) 
Looks good to me. As for your 2 questions, I don't see why not.

vlazar:
Yes, wanted to discuss since it's API :)
Added `text` property to both `awesomplete-selectcomplete` and `awesomplete-highlight`.

PR #16868, opened 2016-03-13
Make awesomplete instance awailable on input element

vlazar:
Having Awesomplete instance available in events as `this.awesompelete` will handy.
Discussed here https://github.com/LeaVerou/awesomplete/issues/16821#issuecomment-185706841

``` javascript
// in all event handlers:
var me = this.awesomplete;

me.ul.children[me.index]; // selected <li> DOM node
me.suggestions[me.index]; // selected suggestion item

// and so on...
```

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -103,7 +103,7 @@ var _ = function (input, o) {
 		this.list = this.input.getAttribute("data-list") || o.list || [];
 	}
 
-	_.all.push(this);
+	_.all.push(this.input.awesomplete = this);
 };
 
 _.prototype = {


LeaVerou:
How about a `WeakMap` on `Awesomplete`? Then to get the instance you'd do something like `Awesomplete.inputs.get(element)`. Then we don't need to add properties to elements, which many people have a knee jerk reaction against.
Also, if we do end up making Bliss a dependency, we can add it to `element._`.

vlazar:
We've discussed WeakMap before. I thought it wasn't an option for the nearest future with current support in browsers. https://github.com/LeaVerou/awesomplete/issues/16821#issuecomment-186079588

LeaVerou:
As long as it's not used internally, people can always choose whether they want to use it or not and take the reduced browser support (or include a polyfill).
I’m not against having auxillary features with different browser support if it's clearly indicated in the docs.
We would, however, need to check for it before using it, to avoid breakage in environments where it's not supported.

PR #16872, opened 2016-03-16
Fire awesomplete-highlight event only if some item was really highlighted

vlazar:
When you move through the list with down/up keys `awesomplete-highlight` is now triggered even after you reach the latest item and press down key again or you reach the first item and press up key again.

In those 2 cases no item became highlighted, but `awesomplete-highlight` event is fired anyway. Is it a desired behavior or is it a bug?

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -188,11 +188,11 @@ _.prototype = {
 		if (i > -1 && lis.length > 0) {
 			lis[i].setAttribute("aria-selected", "true");
 			this.status.textContent = lis[i].textContent;
-		}
 
-		$.fire(this.input, "awesomplete-highlight", {
-			text: this.suggestions[this.index]
-		});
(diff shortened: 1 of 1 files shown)


awesomplete.js
```diff
@@ -188,11 +188,11 @@ _.prototype = {
 		if (i > -1 && lis.length > 0) {
 			lis[i].setAttribute("aria-selected", "true");
 			this.status.textContent = lis[i].textContent;
-		}
 
-		$.fire(this.input, "awesomplete-highlight", {
-			text: this.suggestions[this.index]
```
vlazar, line 194: After the end/before the beginning, when `this.index = -1` the text would be `undefined`.

LeaVerou:
Huh, no idea why it was like that, even though it was obviously intentional (otherwise I'd have put it in the conditional). Anyway, merged!
Thanks for spotting this!

PR #16878, opened 2016-03-29
Allow user to specify pre-existing container element

fractaledmind:
This is similar to ##16805, but in-line with the code style and API design of #16787. It is the pair of #16787. 

Here we allow the `container` object to be set using a selector. This also changes the default className for the container [1] to make things a bit more explicit and [2] to make the backwards-compatible CSS possible.


LeaVerou:
Whoa, I think you accidentally included a bunch of unrelated commits?

Thanks for sending a PR!

PR #16879, opened 2016-03-29
Allow for tab-completion, if not autoFirst and not shift+tab

fractaledmind:
A fixed version of #16833.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -60,9 +60,9 @@ var _ = function (input, o) {
 			var c = evt.keyCode;
 
 			// If the dropdown `ul` is in view, then act on keydown for the following keys:
-			// Enter / Esc / Up / Down
+			// Enter / Tab / Esc / Up / Down
 			if(me.opened) {
-				if (c === 13 && me.selected) { // Enter
+				if ((c === 13 || (c === 9 && !evt.shiftKey && !me.autoFirst)) && me.selected) { // Enter or Tab
 					evt.preventDefault();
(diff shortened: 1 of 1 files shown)


LeaVerou:
LGTM. @vlazar, thoughts?

vlazar:
Sorry @LeaVerou, a bit busy now to carefully think about it. Some thoughts though. More on original #16833 and expected behavior.

If I understood correctly, the  idea was to be able to select currently highlighted item by `Tab` key. So that instead of pressing `Enter` and then `Tab` you only need to press `Tab`.

So if `Tab` now selects an item and goes to next filed, why shouldn't `Shift+Tab` do the similar: select an item and go to previous field?

I'm also curious, if the behavior in this PR would be expected by default by all users. Any demos for autocomplete controls behaving in similar way?

Also, the check for `autoFirst` option disables tab completion. Won't this confuse people?

vlazar:
Code itself looks good, but no tests :( With 3 new branches there should up to 8 new tests ;)

In case of LGTM for behavior on this PR I can help with tests.

Nepherim:
@vlazar:jQuery UI uses tab to select, also shift-tab selects. https://jqueryui.com/autocomplete/

vlazar:
Thanks @Nepherim! So in JQuery UI `Tab` and `Shift+Tab` at least work consistently.

Also, the `autoFocus` option does not change the tab behavior. See http://jsfiddle.net/Bqujj/

PR #16880, opened 2016-03-30
Remove the aria-selected=true attribute from any LIs on hover.

fractaledmind:
This is a feature PR or a bug-fix PR, depending on how you think about the current behavior I guess.

Currently, you can use the keyboard to navigate the LIs in the dropdown, marking the selected one as `aria-selected=true`. With one LI set like so via the keyboard, you could also hover over another LI:

example

This PR ensures there is only ever one item being 'highlighted' by removing the `aria-selected=true` attributes. Currently, it removes that attribute, but does not change `this.index`. I'm not sure what the preferred behavior would be. There are a few options:
1. Set `aria-selected=true` for the item being hovered over and set `this.index` to index of item being hovered over.
2. Have no `aria-selected=true` items when using mouse, always reset `this.index` to -1.
3. Have no `aria-selected=true` items when using mouse, but leave `this.index` set to whatever it was (so hitting down arrow again would select the next item in the list, leaving off from where you where before you hovered over something).
4. Leave it as it is currently.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -79,21 +79,29 @@ var _ = function (input, o) {
 
 	$.bind(this.input.form, {"submit": this.close.bind(this)});
 
-	$.bind(this.ul, {"mousedown": function(evt) {
-		var li = evt.target;
+	$.bind(this.ul, {
+		"mousedown": function(evt) {
+			var li = evt.target;
 
-		if (li !== this) {
+			if (li !== this) {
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hi there, thanks for the contribution! I'm a bit confused though and the diff is hard to parse. 
So which element has `aria-selected="true"` now? The hovered one? Or none?

Issue #16886, opened 2016-04-27
Highlighting not displayed on extra values when using "Multiple values"


tdruez:
Hi Lea,

I've noticed an issue using the "Multiple values" example.
The highlighting on the suggestions works fine on the first item, but no anymore on the second item, thrid, and so on...
Reproducible using the input at http://leaverou.github.io/awesomplete/#multiple-values

Do you think this is something we can solve directly when creating the Awesomplete instance, or does this required a change in the library?

Thanks for Awesomplete by the way, it's an excellent piece of software :)

vlazar:
Related #16814 (examples enhancement)

vlazar:
Absolutely doable.

You need to define a custom `item` function, which is called to render each item. Original `item` function performs match on full input value. In Multiple values example you need to match against the last word in comma-separated list.

teqnology:
Adding here the fix for this issue.
When adding your own `new Awesomplete('input[data-multiple]' ...`
just add the `item` function along with the `filter` and `replace`:

```
filter: function (text, input) {
    ...
}
item: function (text, input) {
    var s = input.split(",");
    var lastItem = s[s.length-1];
    var html = lastItem.trim() === '' ? text : text.replace(RegExp(Awesomplete.$.regExpEscape(lastItem.trim()), "gi"), "<mark>$&</mark>");
    return Awesomplete.$.create("li", {
        innerHTML: html,
        "aria-selected": "false"
    });
},
replace: function (text) {
    ...
}
```

tdruez:
Fixed using the `item` function from https://leaverou.github.io/awesomplete/#multiple-values

PR #16890, opened 2016-05-11
Highlighting extra values when using "multiple values" #16886

jadet:
Proposed fix for #16886.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -276,6 +276,11 @@ _.SORT_BYLENGTH = function (a, b) {
 };
 
 _.ITEM = function (text, input) {
+	if (input.indexOf(",")>1){
+		var inputs=input.split(',');
+		input=inputs.slice(-1)[0];
+	}
+	
 	var html = input === '' ? text : text.replace(RegExp($.regExpEscape(input.trim()), "gi"), "<mark>$&</mark>");
 	return $.create("li", {
 		innerHTML: html,


vlazar:
Thanks for contribution @jadet!

However I'm not sure, if this special case should be handled directly in default `item` function. It is very generic now like default `filter` and `replace` which are not handling anything related to Multiple values example either.

The theme of Awesomplete is to be lightweight, while at the same time be customizable. So default functions are written to be very generic.

What I was referring to in this comment https://github.com/LeaVerou/awesomplete/issues/16886#issuecomment-215444117 is that in Multiple values example in advanced examples here http://leaverou.github.io/awesomplete/#advanced-examples can be fixed by adding a custom `item` function to the example.

On the other hand the need to provide 3 functions with related regexes just to handle a case with comma separated input is not very simple to understand. Fixing a bug in examples will be a good start and will push us to simplify API, so that later similar examples can become easier to implement.

vlazar:
@LeaVerou what do you think about handling this in core?

LeaVerou:
I'm opposed to merging this PR, it seems like it's handling a very special case in the core.

I would be open to discussing ways to make this use case easier though. Any suggestions, @vlazar?

vlazar:
I was thinking that the `filter`, `item` and `replace` functions in multiple values example need different, but related parts of original  input.

Let's say we've entered `CSS, JavaScript, ht` in the input field.

The `filter` function only needs `ht`part to do it's job (find `HTML` suggestion item), and if it gets this part from outside somehow, there is no need to provide this function at all.

The `item` function only needs to be provided to fix a bug with highlighting. It also only needs `ht` part from original input, and if it gets this part from outside somehow, there is no need to provide this function either.

The `replace` function needs the `CSS, JavaScript,` part, so that it can combine `CSS, JavaScript,` and `HTML` (selected item found by `ht`) into final `CSS, JavaScript, HTML,`. This function should be provided anyway, but the thing that splits original input value `CSS, JavaScript, ht` into different parts can be external.

We also have tickets about making Awesomplete to support select-like behavior out of the box. And tags example is a multi-value select in a way. There was also a ticket about knowing a list of already selected items, so that they can be filtered out of autocomplete.

I feel like thinking about all that cases may lead to some useful change in the core to simplify advanced examples.

PR #16894, opened 2016-06-04
update init to allow single item lists

kevinhughes27:
I have a use case where I am templating user data into Awesomeplete and sometimes they only have one item so far. When I join an array of one thing into a csv there is no comma so Awesomeplete doesn't  init correctly.

I added an extra init pathway for strings without comma's that just creates an array out of this single item.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -115,11 +115,11 @@ _.prototype = {
 				this._list = list.split(/\s*,\s*/);
 		}
 		else { // Element or CSS selector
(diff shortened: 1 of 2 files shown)


kevinhughes27:
hmm the more I look at how this whole thing works I realise I should just use a list element to init.

vlazar:
Yes, it's probably too much effort to special case initialization from `data-` attribute for single item list since `data-` attribute supports CSS selectors too.

Issue #16897, opened 2016-06-15
Update input text on autocomplete option highlight


jimf:
Just want to start out by saying that I'd be happy to implement any work that might come out of this issue. I'm more looking for a general +1 on the idea, as well as hash out what the api might look like. With that out of the way...

**Goal:** I'm looking to replace a jQuery-UI based autocomplete implementation with Awesomplete. However, for feature parity, I need the input text to update as I navigate through autocomplete choices. More concretely:
1. A list of autocomplete options are shown (e.g., I've typed "ja" and get the choices "java" and "javascript")
2. I navigate through the options
3. As I navigate, the input text updates to reflect the highlighted choice (pressing down once will update the field text to "java" and so on)
   - hitting escape closes the autocomplete dialog and reverts the input text back to the original value ("ja")
   - "selecting" the option works as is already implemented

I'm fairly confident that this can all be achieved with the existing API. Live-updating the input text is fairly trivial with the `awesomplete-highlight` event, and the original value can be cached on `awesomplete-open`. The hacky part comes with the escape functionality, as the code currently makes no distinction between pressing escape and any of the other methods of closing. I _might_ be able to do something with `setTimeout` / `clearTimeout`, but it's definitely not ideal, hence this issue.

So open questions:
- Is there general value in this behavior? Does it fall inline with the goals of this library?
- If yes, would this be best wrapped up in a single constructor option (and if so, what's a good name), or would it be preferred to leave much of this in userland and a) expose an update to the escape/blur functionality to either close in a "close and do nothing" state, or b) alternatively fire off a second, more meaningful event? Or something else entirely?

vlazar:
@jimf Thanks for very detailed explanation.

So how does jQuery UI autocomplete widget handles this case? Does it have option, event or something else?

jimf:
@vlazar this is the default behavior. Afaict, there is actually no direct way to disable it.

vlazar:
Personally I'm leaning towards having additional event and implementing your use case in userland.

> b) alternatively fire off a second, more meaningful event

It will be trivial to add new event here: https://github.com/LeaVerou/awesomplete/blob/gh-pages/awesomplete.js#L70

It looks like hitting escape is the only case when the currently highlighted choice should be reverted on `close()`. In all other cases where `close()` is called and already triggers `awesomplete-close` event should keep currently highlighted.

But then, of course, it's an API and obvious question would be wether we should add one more event or not.

[...]

PR #16898, opened 2016-06-22
Return early from close if not open

jimf:
If the popup is already in a closed state, return early from `Awesomplete.prototype.close` so as to guarantee that when `awesomplete-close` events get triggered, the popup _was_ actually closed.

See discussion in #16897.


LeaVerou:
Thanks!

PR #16929, opened 2016-08-08
Change getter for `opened` property to use instance variable

mrparkers:
**Problem:**  When testing code that relies on Awesomeplete, attempting to stub the constructor using [Sinon](http://sinonjs.org/) yields an error when Sinon attempts to convert the function to a string to display test results.  Here is the relevant Sinon.js code that is causing our test suite to break:

``` js
sinon.functionToString = function toString() {
    if (this.getCall && this.callCount) {
        var thisValue,
            prop;
        var i = this.callCount;

        while (i--) {
            thisValue = this.getCall(i).thisValue;

            for (prop in thisValue) {
                if (thisValue[prop] === this) { //fails here
                    return prop;
                }
            }
        }
    }

    return this.displayName || "sinon fake";
};
```

The failure occurs at `if (thisValue[prop] === this)`.  When iterating over all of the constructor's properties, it fails at `opened` because the getter relies on `this.ul.hasAttribute()`, and `this.ul` is undefined when stubbing the constructor.  I think that Sinon is making the assumption that getters wouldn't rely on the existence or values of other unrelated properties/functions, which I would agree with.

**Solution:** Create a new instance variable, `isOpened`, that is modified whenever `open()` or `close()` is called.  Return this variable instead of relying on `this.ul.hasAttribute()`.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -12,6 +12,8 @@ var _ = function (input, o) {
 
 	// Setup
 
+	this.isOpened = false;
+
 	this.input = $(input);
 	this.input.setAttribute("autocomplete", "off");
 	this.input.setAttribute("aria-autocomplete", "list");
@@ -143,7 +145,7 @@ _.prototype = {
 	},
 
 	get opened() {
-		return !this.ul.hasAttribute("hidden");
+		return this.isOpened;
 	},
 
 	close: function (o) {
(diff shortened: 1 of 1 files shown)


mrparkers:
@LeaVerou I'm sorry to bother you, but would you mind addressing this PR?  My team is unable to test the parts of our app that consume Awesomplete as it is, so we're forced into pointing to my fork as our dependency, which is not a fantastic long-term solution.  It would be really nice for this to be included in an official release.

Please let me know if you need anything else from me before this can be merged.  If needed, I'd be happy to throw together an example repo that demonstrates the problem we're having.

LeaVerou:
Merged, thanks.

PR #16930, opened 2016-08-16
Loop around options using: next or previous

dhoko:
Hi,

A small fix if you select the previous/next option. 
## Before

before
## After

after

Case 1:
Auto select the first item of the list when you press the DOWN_KEY (_at the end of the list_).

Case 2:
Auto select the last item of the list when you press the UP_KEY (_at the top of the list_).


LeaVerou:
Thanks for contributing!

This behavior is intentional, as an escape mechanism. But if others want this to be merged, fine by me. No strong opinions here :)

dhoko:
Thx :) 
IMHO, it feels weird to lost the focus.
If we use a datalist, it's the same behaviour as this PR. cf:

output

awesomplete.js
```diff
@@ -169,14 +169,15 @@ _.prototype = {
 
 	next: function () {
 		var count = this.ul.children.length;
-
-		this.goto(this.index < count - 1? this.index + 1 : -1);
+		var defaultPos = count ? 0 : -1;
+		this.goto(this.index < count - 1? this.index + 1 : defaultPos);
```
LeaVerou: No need for a variable here, please put it inline?

dhoko: I can :) 
_imho_ I don't think it's a good idea, because a ternary inside another one is not easy to read.

LeaVerou: Not even with parens? :)

dhoko: Done (_but imho still not as readable as before_).

dhoko:
No changes with the .min after a new build :)

LeaVerou:
Ok, merged! Thanks!

dhoko:
:ok_hand: 

Are you going to deploy a new version (_patch ?_) via npm/bower ?

LeaVerou:
Not just for this, but yes once there are enough changes that warrant a new version :)

ysndmr:
Although I have made all the updates, it still doesn't open. I couldn't fix this problem.  is there someone who can help?

PR #16963, opened 2016-09-08
Feature: Sourcemaps

marcobiedermann:
Add `gulp-sourcemaps` to create a CSS sourcemap.
This might be helpful for debugging.
Unfortunately `gulp-closure-compiler-service` does not support the usage of sourcemaps.
Maybe you could switch to [gulp-uglify](https://www.npmjs.com/package/gulp-uglify) which does support it and is more active in development.
Using `gulp-uglify` is a bit less efficient (35 bytes) but I think it is worth it to support sourcemaps.
What do you think Lea?


LeaVerou:
Sure, but you'd have to make that change too then :)

Thanks!

marcobiedermann:
@LeaVerou Done!

To sum it up:
- add Sourcemaps for CSS and JavaScript files
- replace `gulp-closure-compiler-service` with `gulp-uglify`

LeaVerou:
Awesome, thanks!!

marcobiedermann:
@LeaVerou You are very welcome. I have to thank you for this awesome library 😄 🎉

Issue #16976, opened 2016-09-22
Save extra info about each list item


jandramila:
So i´ll first describe my use case: Im using awesomeplete with google places API, so basically, im populating the _list collection for the awesomplete on the location input with an async call to google´s API, for each item i have plenty of information such as geometry and a google_place_id which identifies the given place.

![locationexample](https://cloud.githubusercontent.com/assets/10802091/18757954/6531a0a2-80cc-11e6-94c6-06030c1c4208.gif)

I't would be nice if i would be able to retrieve that data from the autocomplete selection, so whenever the awesomplete-selectcomplete event is fired, the data attached to the given item is available.

I will mostly sure add this and use a customized version of the library but i can add this to a PR if anybody likes the idea.

jandramila:
LoL i never saw this #16938 my bad

vlazar:
Currently we support a separate `label` and `value` for each suggestion (the original array of strings, objects or arrays). See docs http://leaverou.github.io/awesomplete/#basic-usage

Additional properties are not directly possible due to API backward compatibility shim we use right now https://github.com/LeaVerou/awesomplete/blob/b757022a728a9f825c545f2235b99c5e98a85874/awesomplete.js#L302-L315. See explanation here  https://github.com/LeaVerou/awesomplete/issues/16888#issuecomment-218536654

But this would be a really nice feature to have.

El4a:
@jandramila so did you work out a customized solution? I'm facing the same problem right now so I'd love to hear your solution (I can't work it out :( )

jandramila:
@El4a Yeah, i achieved it, but it only works when you programmatically set list items, i just added extraData to these lines:

```javascript
...
// Private functions

function Suggestion(data) {
	var o = Array.isArray(data)
	  ? { label: data[0], value: data[1] , extraData: data[2]}
	  : typeof data === "object" && "label" in data && "value" in data ? data : { label: data, value: data };

	this.label = o.label || o.value;
	this.value = o.value;
	this.extraData = o.extraData;
}
Object.defineProperty(Suggestion.prototype = Object.create(String.prototype), "length", {
	get: function() { return this.label.length; }
});
Suggestion.prototype.toString = Suggestion.prototype.valueOf = function () {
	return "" + this.label;
};
...
```
I´m using my own awesomplete customized version on a react app, you can refer the list via jQuery selector or whatever, in my case, i use a component ref:

```javascript
[...]

Issue #16938, opened 2016-08-25
Access Original List Item?


joshvickerson:
I'm having trouble figuring out if I can access the original list item. My goal is to set a data-attribute on the generated list item so I can access the value when the item is selected. The stored value is important to a function call I want to make when something is selected.

My data list looks something like this:

``` html
<ul id="myList">
  <li data-thing="value">Thing1</li>
  <li data-thing="value">Thing2</li>
  <li data-thing="value">Thing3</li>
</ul>
```

And I initialize Awesomplete like this:

``` javascript
var input = document.getElementById("myInput");
var ac = new Awesomplete(input, {
    list: "#myList",
    minChars: 1,
    item: function(text, input) {
        // I want to set a data-attribute on the generated item
        // by reading a data-attribute from the original list item
    }
});
```

I cannot figure out how to access the original list item to do so. Am I on the right track here, or is there a better way to go about this?

**Edit: I noticed in the documentation that the data function has two parameters, item and text, where item is the original list item. I tried testing this with a console log like so:**

``` javascript
data: function(item, input) {
    console.log(item);
}
```

**But what is logged is just an object with a label and value. I thought that "original list item" meant a reference to the element, so I tried `item.getAttribute('data-thing')` but getAttribute is undefined. Am I misunderstanding something?**

Ruff9:
I have kind of the same problem, I think maybe a few examples of how you use events could do the trick. 

I want to access to the list item when the awesomplete-select event is fired. I'm sorry, but I don't get what you mean by :

`Callback will be passed an object with text (selected suggestion) and origin (DOM element) properties.`

Ruff9:
I solved my pb, maybe that can help @joshvickerson . At first i catched the event with jquery#on.

`$input.on('awesomplete-select', function(e) {
  console.info('A selection was made');
  console.log(e.text.label); 
});`

And it failed to give me the info. But when I did this one, it worked:

`input.addEventListener('awesomplete-select', (e) => {
  console.info('A selection was made');
  console.log(e.text.label); 
});`

joshvickerson:
I'm not quite sure what you're suggesting, @Ruff9, but what I ended up doing is a bit of a hack...

I was able to initialize my list as an array instead of using HTML. It's an array of objects in this format:

``` javascript
{
    label: "the text I want displayed",
    value: "the value I was trying to put in a data-attribute"
}
```

I then create my Awesomplete list like this:

``` javascript
[...]

Issue #16888, opened 2016-04-29
How to customize dropdown list (add images, etc)


rightaway:
In the dropdown list of suggestions, instead of just text appearing is it possible to have images next to the text? For example, if the dropdown contains a list of food, there could be a small picture of that food item next to it. Is it possible to do or could this feature be added?

LeaVerou:
It is, as of v1.1!! @vlazar, could you explain how, since you added that feature?

vlazar:
It may be possible, since we have separated `label` and `value` for each suggestion item in v1.1.

List of items provided via `list` property can be any object with properties like `{ label: ..., value: ..., url: ...}`. 

However, since we also support string based v1.0 API we wrap items internally with special `Suggestion` object. This object has only `label` and `value` properties from original item on the list (no `url`). So right now you can't just use `url` property in functions like `item` and others.

As workaround, if you only need text and image, you can use `value` for text and `label` for image url and define custom `item` function to show image in each suggestion.

@LeaVerou I think we should definitely try to extend Awesomplete for this case, so that any property in original list will be accessible in custom functions.

TxHawks:
+1 for making all properties accessible

rightaway:
@vlazar, thanks a lot, could you also provide a simple example of how to use the properties you just mentioned? I've tried it but I don't think I'm doing it right...

vlazar:
@rightaway The following options should work for you then:

``` javascript
list: [ { value: 'Text for dropdown and input field', label: 'http://example.com/img.jpg' }, ...],

item: function (suggestion, input) {
    // notice the use of suggestion.value here, default function effectively uses suggestion.label
    var html = input === '' ? suggestion.value : suggestion.value.replace(RegExp($.regExpEscape(input), "gi"), "<mark>$&</mark>");

    // here you will have image URL in suggestion.label property, so you can add it to LI
    return $.create("li", {
        innerHTML: html,
        "aria-selected": "false"
    });
},

filter: function(suggestion, input) {
    // the default function uses suggestion.label here as well, so we need to override this too
    return Awesomplete.FILTER_STARTSWITH(suggestion.value, input);
}
```

Insood:
Hey guys, thanks for the script. This is what I ended up doing for my purposes:

```
AWESOMPLETES = {}

image_search_input_change = (e) ->
    element = this
    $.getJSON( "/images", {file: this.value}).success( (data) ->
        AWESOMPLETES[element].list = data
    )

[...]

PR #17001, opened 2016-11-07
Spread list props to suggestion

Secretmapper:
Spread all props of list object to Suggestions object.

Super direct way and not sure if it breaks backwards compat/works in all browsers

Adresses https://github.com/LeaVerou/awesomplete/issues/16976, https://github.com/LeaVerou/awesomplete/issues/16938, https://github.com/LeaVerou/awesomplete/issues/16888

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -304,6 +304,8 @@ function Suggestion(data) {
 	  ? { label: data[0], value: data[1] }
 	  : typeof data === "object" && "label" in data && "value" in data ? data : { label: data, value: data };
 
+	for (var attr in data) { this[attr] = data[attr]; }
+	
 	this.label = o.label || o.value;
 	this.value = o.value;
 }


vlazar:
Thanks for your contribution!

I think Array and non label/value Object cases can be skipped. No need to copy properties when `data` is Array or Object without label/value properties.

Also we shouldn't probably copy all properties from objects. We [inherited Suggestion object from String and re-defined some properties](https://github.com/Secretmapper/awesomplete/blob/1c2ecc3393c9a7103b24ed880008883b92217898/awesomplete.js#L312-L317). Probably not all String features currently work for Suggestion, but overriding all properties on Suggestion blindly can probably lead to more potential issues.

@LeaVerou cares a lot about backward API compatibility. We should not break Awesomplete API.

Secretmapper:
@vlazar Good point. I didn't even realize Suggestion inherits from String.

Perhaps a better solution then is to namespace the passed object as a prop, ie `this.data = data` and the object itself can be accessed from the suggestion. What do you guys think?

vlazar:
I was thinking about having `suggestion.data` for all other props as a way to not complicate implementation, but then API becomes weird:

``` javascript
suggestion.label
suggestion.value
suggestion.data.everythingElse
```

Of course you can just use:

``` javascript
suggestion.data.label
suggestion.data.value
suggestion.data.everythingElse
```

But once introduced, this API would probably have to stay forever.

Lea, what do you think?

TxHawks:
+1 for

``` js
suggestion.data.label
suggestion.data.value
suggestion.data.whatever
```

joonassandell:
Any updates on this?

vlazar:
@LeaVerou What do you think?

LeaVerou:
> @LeaVerou cares a lot about backward API compatibility. We should not break Awesomplete API.

To be clear, it's not backwards compatibility I care so much about, it's usability. Awesomplete's main use case is people using it with a list of strings. Allowing more complex use cases should NEVER complicate the simple, common case. That's how people end up with terrible APIs that are a pain to use: They design the API based on the most complicated cases they want to support, and its gets in the way of the majority of their users with simpler use cases. Good APIs make the complex possible and the simple easy, not just one of the two.

That said, I can't see how this PR could break anything. Unless the properties of the object passed in happen to have the same name as string methods that are used, there shouldn't be any issue.

I would have some stylistic comments though:
1. Please follow the style of the rest of the code, with blocks being on separate lines even if they only have one line.
2. Please rename `attr` to `prop` or (better) `property`. These are not attributes, they are properties. Attributes can be confused with HTML attributes.
3. As @vlazar mentioned, you don't need this unless `data` is an object. Since there is another check about whether this is the case, we should convert it to a proper block if and do it in there.

vlazar:
@LeaVerou How about this idea? https://github.com/LeaVerou/awesomplete/pull/17001#issuecomment-259620437

LeaVerou:
> @LeaVerou How about this idea? #17001 (comment)

The indirection seems like a hassle IMO. The chances of *both* having a conflict *and* people trying to use the String method/property that is obscured by the data are pretty slim if you ask me, and easy to fix by just renaming the property in their data.

Lesha-spr:
Any updates on this?

awesomplete.js
```diff
@@ -304,6 +304,8 @@ function Suggestion(data) {
 	  ? { label: data[0], value: data[1] }
 	  : typeof data === "object" && "label" in data && "value" in data ? data : { label: data, value: data };
 
+	for (var attr in data) { this[attr] = data[attr]; }
```
bholtbholt, line 307: - [x] Please follow the style of the rest of the code, with blocks being on separate lines even if they only have one line.
- [x] Please rename attr to prop or (better) property. These are not attributes, they are properties. Attributes can be confused with HTML attributes.

```suggestion
	if (typeof data === "object") {
		for (var property in data) {
			this[property] = data[property];
		}
	}
```

awesomplete.js
```diff
@@ -304,6 +304,8 @@ function Suggestion(data) {
 	  ? { label: data[0], value: data[1] }
 	  : typeof data === "object" && "label" in data && "value" in data ? data : { label: data, value: data };
```
bholtbholt, line 305: - [x] As @vlazar mentioned, you don't need this unless data is an object. Since there is another check about whether this is the case, we should convert it to a proper block if and do it in there.

```suggestion
	if (Array.isArray(data)) {
		o = { label: data[0], value: data[1] }
	} else if (typeof data === "object") {
		for (var property in data) {
			this[property] = data[property];
		}
	} else if (typeof data === "object" && !("label" in data && "value" in data)) {
		o = { label: data, value: data }
	}
```

bholtbholt (commented):
@LeaVerou if this is something still worth adding I'm happy to take it over the finish line. It fits a use-case that's coming up for me!

I think I'll need to open a new PR, but I've put comment suggestions here as well.

PR #17006, opened 2016-11-15
fix: allow selection on mobile ios

### LINKED ISSUE #16712: Touch on completion list not firing on iOS

Narretz:
On mobile ios up until ios9, no mousedown event is fired, because ios keeps the focus
on the currently active input element. However, it does fire a click event.

To work around this problem, but keep the current behavior, we add a click event
handler to the existing mousedown handler, which will now only preventDefault,
which keeps blur from firing, but allows the click handler to do the actual selection.

This new behavior additionally means that clicking and holding (dragging) the mouse away
from the autocomplete list will not result in a selection, mimicking the normal
mouse behavior.

Unit tests pass in Chrome 54
Manually tested on Win 10 Chrome 54, FF 49, Edge, IE11; on ios with 9.3.5 and 10.1
FF 

This is an alternative to #16715 and fixes #16712

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -91,7 +91,14 @@ var _ = function (input, o) {
 			"submit": this.close.bind(this, { reason: "submit" })
 		},
 		ul: {
+			// Prevent the default mousedowm, which ensures the input is not blurred.
+			// The actual selection will happen on click. This also ensures dragging the
+			// cursor away from the list item will cancel the selection
 			"mousedown": function(evt) {
+				evt.preventDefault();
+			},
+			// The click event is fired even if the corresponding mousedown event has called preventDefault
+			"click": function(evt) {
 				var li = evt.target;
 
 				if (li !== this) {
--- a/test/events/clickSpec.js
+++ b/test/events/clickSpec.js
@@ -1,0 +1,61 @@
+describe("click event", function () {
+
+	$.fixture("plain");
+
+	subject(function () {
+		return new Awesomplete("#plain", { list: ["item1", "item2", "item3"] });
+	});
+
+	def("li", function () { return this.subject.ul.children[1] });
+
+	beforeEach(function () {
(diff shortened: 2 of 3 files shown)


MHanslo:
Thank you @Narretz, I gave it a go and this does indeed fix the issue I'm having. Good work.

Narretz:
@MHanslo That's good to hear, thanks for testing. Follow up question, did you test it in browser or inside an app (e.g. cordova)? I only tested in browser but as I found out the problem was actually reported for an app.

MHanslo:
@Narretz, sorry, I've been quite busy lately. I recall I tested this in the browser; using the Chrome iPhone emulator I could always replicate the touch bug on our app. Dropping in your fix corrected the bug; I could actually perform a selection.

Narretz:
@LeaVerou Just stumbled over this again. Are you still maintaining this library?

aabluedragon:
Is it going to be merged anytime soon?

LeaVerou:
I’m fine merging this if the conflicts are resolved.

In general, if you really want a PR merged, merge it locally, test it out, and report back. That's much more effective than whining or asking if a library is still maintained. But that takes more time you see…

Narretz:
@LeaVerou I've rebased the PR

LeaVerou:
Merged, thanks!

PR #17025, opened 2016-12-10
Scroll to highlighted element when list container height is fixed

netchampfaris:
When the `ul` is of fixed height, navigating with keyboard does not scroll to selected element.
This PR introduces scroll to element functionality. Hope it is according to coding standards.

awesomplete

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -197,6 +197,9 @@ _.prototype = {
 			lis[i].setAttribute("aria-selected", "true");
 			this.status.textContent = lis[i].textContent;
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thanks for contributing! LGTM, have you tested this with dropdowns that don't scroll?

netchampfaris:
@LeaVerou 
Yes, it works. The scroll position calculates to `0` which means it doesn't require scrolling.

LeaVerou:
Merged, thanks!

Issue #16887, opened 2016-04-29
Timeout before filtering >10k rows


nixprosoft:
When datalist contains more than **10k** rows, **browser is laggin** while making quick input from barcode scanner emulating keyboard key press. Input contains about 26 latters. 
Is it possible to optionaly make a tiny pause before start fintering?

vlazar:
There is no out of the box support for delay currently.

Meanwhile, why don't you just assign a final value from barcode scanner instead of emulating each letter keypress event?

nixprosoft:
It's native function of the scanner, it's all about driver. I can't change that behaviour.
Maybe awesomplete's custom filter function can help?

vlazar:
Filter function just filters data.

Flow:
1. Keypress in input: https://github.com/LeaVerou/awesomplete/blob/gh-pages/awesomplete.js#L57
2. Evaluate call https://github.com/LeaVerou/awesomplete/blob/gh-pages/awesomplete.js#L223
3. Filter call https://github.com/LeaVerou/awesomplete/blob/gh-pages/awesomplete.js#L236

There is also `minChars` option and check here https://github.com/LeaVerou/awesomplete/blob/gh-pages/awesomplete.js#L227

If the barcode is always approximately 26 letters, you can try to set `minChars` option, so that filter won't run until specified number of characters are entered in field.

nixprosoft:
Can't use minChars in that way, because same input field is used for manual input (by hand), and minChars is set to 3.
Do you think I can solve the problem with filter?

Issue #16832, opened 2016-01-29
tab to complete?


Nigelli:
Hi, not necessarily an issue but as a user I expect to be able to select an item from the list by hitting the tab key.

Easy to fix by adding key code 9 to line 72 in awesomeplete.js?

`if ((c === 13 || c === 9) && me.selected) { // Enter or Tab`

TxHawks:
That's not how a `select` element works...

`tab` moves you away from the widget, the arrow keys are used for selection

Nigelli:
Yes of course, but once you have you choice selected if you tab to move away should this choice not be chosen?

Isn't it logical to assume that if the user has made the effort to go down the list and highlight the most appropriate item then it is unlikely their intention when tabbing away will be to cancel their choice?

LeaVerou:
> Yes of course, but once you have you choice selected if you tab to move away should this choice not be chosen?
> 
> Isn't it logical to assume that if the user has made the effort to go down the list and highlight the most appropriate item then it is unlikely their intention when tabbing away will be to cancel their choice?

That's a very good point. I'm convinced. We should not `preventDefault()` on tab, but it should still make a selection.

LeaVerou:
Although it would be a problem if the dropdown opens automatically, so maybe not then.

Nigelli:
Sorry I might be missing something obvious but i'm not sure what you mean? If autoFirst is set to false and the user tabs nothing will be selected and therefore, tab will just move the user away.

If autoFirst is true then yes tabbing away would select the first option in the list. But isn't that the point of autoFirst? Using escape would still cancel this.

LeaVerou:
Yes, I meant when using `autoFirst` as well.

Frazer:
using tab to select and then stay in the field is not quite right either,  may I suggest:

    if (c === 13 && me.selected) { // Enter
        evt.preventDefault();
        me.select();
    }else if ( c === 9 && me.selected) { //Tab - dont prevent default
        me.select();
    }else....

kball:
Just ran into this as well - looks like there was some movement in #17032 but it got tangled up with a lot of other changes and didn't happen. @LeaVerou would you accept a targeted PR that implements just this behavior behind an option flag?

LeaVerou:
@kball If the code is reasonably good & compact, sure!

kball:
@LeaVerou: Submitted in #17145. Came out pretty clean & compact, but let me know if you think anything needs fixing and I'll happily update

[...]

PR #17032, opened 2016-12-21
Implementes filter delay and adds a few keybindings

andrejborstnik:
### Disclaimer
While working on a project I found myself needing a few expansions to the ```awesomplete```.  Since I saw a few other people were asking for similar features I submit this small contribution. I invite you to comment on any feature.
### Changes
* The Up/Down arrows now open the ```ul```. Such a feature is needed if we want complete (and convenient) mouse-less control. E.g. I navigate (with Tab) to an ```input``` with something already written and want to see recommendations. Or I have just clicked Escape, but don't want to change the text to get the ```ul``` back.
* The filtered list remembers where we left off (we just don't set ```index = -1``` on close). It seems only natural to me to continue highlighting where we left off last time, if the filtered list has not changed.
* Added the Tab and Shift-Tab keybindings. For me tabbing while inside the ```ul``` is natural. Regarding the concern about navigating inputs with Tab: Tab does not open the ```ul```, so by selecting a value (enter), or by closing the ```ul``` (Escape) we are free to navigate away. I also considered #16832, but I never use Tab to select like that - it only seems natural if the completed text is shadowed in the input field, or there is only one choice (as the command line). But if your experience is different it is an easy fix.
* Added the option to block filtering for a certain amount of time (after each filter), which is necessary for large lists (```100k+```). The default ```0.5s``` seems feasible to me, but feel free to change it to 0 and it will not affect the operation of ```awesomplete```. Also solves #16887.
### Tests
I updated the tests to reflect the changes. I did not (yet) write tests for delayed filtering, but I have tested it. Existing tests pass for ```delay = 0```, but would need to be adapted for other delays, as not every call of the (old) evaluate function is (fully) executed.


LeaVerou:
Thank you so much for this contribution @andrejborstnik, this looks great! What do the others think? Should we merge?

Small note: In the future, it would be much easier for us if you opened separate PRs for separate improvements. It would also mean your PRs get merged faster, because reviewing them is easier.

LeaVerou:
Since this modifies the keyboard UX quite a bit, I would love to try it out and see how it feels (I'm a heavy keyboard user too). Is there any online demo? I tried https://andrejborstnik.github.io/awesomplete/ but it doesn't seem to have the improvements (e.g. pressing down arrow on an input does nothing).

awesomplete.js
```diff
@@ -236,6 +248,25 @@ _.prototype = {
...
+					setTimeout(function () {
+							return this.evaluate();
+						}.bind(this),
+						this.filterDelay);
+					this.filterDelayed = true;
+				}
+				return;
+			}
+			this.lastFilterTime = time;
+			this.filterDelayed = false;
```
LeaVerou: Do we really need a separate flag for this? Wouldn't checking `this.filterDelay > 0)` suffice?

andrejborstnik:
@LeaVerou I will keep multiple PRs in mind next time :). 

The demo on  https://andrejborstnik.github.io/awesomplete/ does have all the improvements, but does not showcase delay. If you want I can write up a demo for that. You can notice the default delay though if you compare it to https://leaverou.github.io/awesomplete/ (delete chars fast and see a delay on my page).

You are not seeing anything on arrow down, because there is nothing in the input - the arrow down only shows recommendations if other conditions are met (e.g. the number of chars written). I programmed it this way, so that the changes to program logic are minimal.

vlazar:
First of all I would 👍 splitting this into separate PRs but we can discuss proposed features first.

Changes with keyboard UX may not be what all users of Awesomplete want. I think we should consider how difficult it would be to add proposed changes with tab/shift-tab without modification of Awesomplete itself versus how difficult it would be to get current behavior with tab/shift-tab if proposed changes are merged in.

Not sure about remembering highlighted item. But again, we need to evaluate it if it's something most users of Awesomplete would expect.

Delayed filtering looks interesting, although personally I would prefer a more generic solution, e.g. adding support for any async suggestion lists. We have a corresponding list of issues https://github.com/LeaVerou/awesomplete/milestone/4

andrejborstnik:
No problem, when we converge to a features list I will create seperate pull requests.

The tab/shift tab could be enabled with an additional option parameter if this feature is not desired by all users - we can than optionally create a poll and see how many people use it. I know I would never expect tab to shift fields while inside the dropdown. But as I stated this was created for my project and is subject to other opinions.

This could again be an option parameter.

The delay was designed for two reasons: to stop unnecessary computation and to improve the UX by reducing browser lag with big lists. 

Async could (if the call only returned the matched values) solve the second problem. But than I only use awesomplete to display, not to autocomplete. 

And async would still not solve the first problem: each time* the input changes we would compute the suggestions. If this is done on the frontend the user might see browser delays and if we do it on the backend we put a lot of unnecessary strain on the server. 

By blocking requests for a period of time we solve both problems at the same time. This is already made optional, as we retain the same functionality by setting delay to 0 (could be default).

PR #17036, opened 2017-01-17
Increase z-index of .awesomplete > ul

BobHarper1:
I found that if the popup menu was rendered over another input in my form (e.g. text input) with a higher z-index, that it was difficult to get the cursor to click on the `<li>` option that appeared to be over that other input. Increasing the z-index is an easy fix.

I think in most cases, as the popup menu appears after keying and disappears on other actions, that UX will be improved by having popup on top of the 'stack', which is why I've used such a high index.


LeaVerou:
I can think of many cases where you **don't** want Awesomplete to be at the top of the stack: Dialog boxes, sticky headers, dragged items etc. I could be persuaded to change it to 2, but 1000 is not only a change that will break many people's websites, but also a bad practice: you are trying to beat another component that has set an unreasonably high z-index, and that's a fight that will never end. Therefore, I'm sorry but I will close this PR, you can always place an extra rule in your stylesheet, though I'd suggest fixing the z-index of the other component. Thanks for contributing though, hope this does not put you off future contributions!

BobHarper1:
No problem, thanks for explaining. I can fix locally and work with that, I guess it something that needs changed as and when.

PR #17058, opened 2017-04-11
Added configurable submitKeys

mirbuf:
Instead of default Enter key, any number of keys can be configured via optional submitKeys: [] array.
Defaults to Comma, Tab and Enter keys.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -10,7 +10,7 @@
 var _ = function (input, o) {
 	var me = this;
 
-  var BACKSPACE = 8;
+  //var BACKSPACE = 8;
   var COMMA = 188;
   var TAB = 9;
   var ENTER = 13;
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -10,6 +10,12 @@
 var _ = function (input, o) {
 	var me = this;
 
+  var BACKSPACE = 8;
+  var COMMA = 188;
+  var TAB = 9;
+  var ENTER = 13;
+  var ESCAPE = 27;
+
 	// Setup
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thank you for working on this! I like the flexibility of being able to specify your own keys, that is very on par with Awesomplete's design principles.
However, I still think Enter should be the default. Tab can interfere with tabbing through form controls, and comma could interfere when your actual options contain commas.

AetherUnbound:
This seems like a great feature, but I agree that having it as default might be problematic. Perhaps it might be made more configurable, i.e. `submitKeys` initially only contains the enter key but there might also be a method to add one's own preferred default keys. 
For instance, we hope to use awesomplete for entering directories. Tabs would be great for that, but perhaps not commas.

mirbuf:
I added submitKeys defaulted that way as a PoC. I agree that could be left just Enter, or Enter + Tab as default keys. And a method to set the new option would be a nice addition. I'll eventually add it, later.

mirbuf:
LeaVerou, I set default submitKeys to [ Enter ].
I also added a method to set user own submitKeys. You can:

setSubmitKeys( KEYCODE )
setSubmitKeys( [ KEYCODE1, ... , KEYCODEn ] )
setSubmitKeys( 'aeiou' )    // this will set 'A', 'E', 'I', 'O', 'U' keys keycodes

Let me know if this may be a more useful addition.

LeaVerou:
Why do we need a function (and such a long one at that)? People can modify the `submitKeys` array, no?

Issue #17060, opened 2017-04-21
Multiple values: Am I to stupid?


Mike-DE-RE:
Hi.

I've tested a lot and learned a lot!

Here is my html-sample:

[test.txt](https://github.com/LeaVerou/awesomplete/files/947536/test.txt)

Here you see what happens, when I select "E10/13, 6V".

![result](https://cloud.githubusercontent.com/assets/27857771/25284703/281eee58-26b8-11e7-9136-a410cd4e080b.png)

As per "Multiple values" sampe I expected a different result:

![sample](https://cloud.githubusercontent.com/assets/27857771/25284706/28328f8a-26b8-11e7-9d3e-e9bd28c9fd75.png)

What do I do wrong?

Thanks

Michael

Mike-DE-RE:
That is very polite of you, not to answer my question. ;-)

Maybe I should have posted this when I started my question. Here is what happens, when I type the first part in Lea's sample:

![temp1](https://cloud.githubusercontent.com/assets/27857771/25373912/b90446f8-299b-11e7-87c9-f8a3501de499.png)

... and here is the same, when I do it in my sample:

![temp](https://cloud.githubusercontent.com/assets/27857771/25373931/d05d16cc-299b-11e7-81ef-2020e7a22f45.png)

It looks to me as if Lea's sample has and "and" condition and mine has an "or" condition.

I have already seen another thread asking about the disappearance of the yellow marking, so I do not post about it here.

Have a lot of fun...

Mike-DE-RE:
You learn something new every day. So here is today's lesson:

![test3](https://cloud.githubusercontent.com/assets/27857771/25419452/1789d88e-2a53-11e7-9ae1-698bd59bd20f.png)

When entering "6V, E10/13" instead of "E10/13, 6V" _multiple values_ work as expected. I am quite sure that it is not intended, because the user will not always know the order of appearance of his search terms. I hope that this showes where the problem comes from.

Can it have to to with the spaces in my strings?

No, it doesn't:

![test4](https://cloud.githubusercontent.com/assets/27857771/25419724/9bf0adc2-2a54-11e7-8bf4-3bccea5093d4.png)

Can it have to do with the commas in my string?

No, it doesn't:

![test6](https://cloud.githubusercontent.com/assets/27857771/25419796/157d9e48-2a55-11e7-8c68-5aa185c54d73.png)

As always, have a lot of fun...

LeaVerou:
Hi there,

[...]

PR #17065, opened 2017-04-28
Hightlight correctly with multiple entries

davidkna:
Fix for one of the issues in #17060.
The new `Awesomplete.data` makes sure only the current part of the input (after the comma) is used.

### DIFF
--- a/index.html
+++ b/index.html
@@ -395,6 +395,10 @@ We can also <strong>generate list items based on user's input</strong>. See E-ma
 		return Awesomplete.FILTER_CONTAINS(text, input.match(/[^,]*$/)[0]);
 	},
 
+	item: function(text, input) {
+		return Awesomplete.ITEM(text, input.match(/[^,]*$/)[0]);
+	},
+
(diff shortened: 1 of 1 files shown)


LeaVerou:
Thanks for working on this!

However, this is a fix that should be performed in the `data-multiple` example in the docs, not in general Awesomplete code. You might want to have entries with commas, and if you're not doing anything special (like the data-multiple example), that should NOT affect your matching.

davidkna:
This PR is for the `data-multiple` example in the docs in the doc I believe. Is `index.html` the right file?

Right now nothing gets `<marked>` if the whole input string does not match the any of the items in the suggestion list. `filter` uses the same regex and handles entries with commas incorrectly as well right now. When this code fails it's likely entries with commas might just be `filter`ed.

The data-multiple sample needs another fix to correctly handle list items with commas but I'm not sure how to best solve that.

LeaVerou:
Oh my bad! I didn't see the filename 😊  Merged now, thanks!!

PR #17068, opened 2017-05-11
trying to meet a11y criteria

stephenkeable:
After comments here -
https://twitter.com/tvararu/status/860549930552840193

Addition of aria-activedescendant to input when list item selected,
with id from newly added id attributes for each item in list
Swap to combobox role for input as better voiceover description
Add aria-owns to connect input to list of results
Modified status text to inform user how many results, along with which
results is selected (including 1 of 4, 2 of 4 etc to indicate position
in list)

Updated tests to match new functionality, feedback on whether these tests are suitable wanted.

Tested behaviour using macOS voiceover, however further testing more
than welcome.


awesomplete.js
```diff
@@ -42,15 +43,18 @@ var _ = function (input, o) {
 
 	this.ul = $.create("ul", {
 		hidden: "hidden",
+        role: "listbox",
+        id: "awescomplete_list",
```
LeaVerou: This won't work very well if you have multiple instances of Awesomplete (unless I'm missing something...)

LeaVerou:
Thank you so much for this, and sorry it took so long to get back to you!
See the comment I left. 
I wish ARIA wasn't so id-focused, it makes anything dynamic a PITA :(

awesomplete.js
stephenkeable: Yes that's true, we ran into this problem in the final testing of - https://github.com/allies-computing/accessible-address-autocomplete

We solved it there by incrementing a count of instances and using that with the ID naming.

Will update this evening on my fork.

stephenkeable: Also think I spelt awesomplete wrong in a few names, so will correct that too :)

awesomplete.js
```diff
@@ -9,14 +9,17 @@
 
 var _ = function (input, o) {
 	var me = this;
+    
+    // Keep track of number of instances for unique IDs
+    Awesomplete.numInstances = (Awesomplete.numInstances || 0) + 1;
```
LeaVerou: Minor, but this name is a bit awkward. What about `size`, `length`, or `count`?

stephenkeable:
Must admit I'm not 100% sure how to solve the ESLint issue with the instance count solution, any advice welcome.

LeaVerou:
What ESLint issue?

LeaVerou:
LGTM, should I merge?

stephenkeable:
The lint issue was about "'Awesomplete' is not defined" however it works and if you are happy with it, I am too.

PR #17073, opened 2017-06-01
Mention destroy method (added by #17067) in API docs

sneakertack:
It was nice to be able to use the destroy method from #17067. However, there's no mention of its existence in the docs... so this PR updates the API methods section to let readers know that it's available.


LeaVerou:
Thanks!

PR #17075, opened 2017-06-04
fix list of programming languages

Icewild:
Fixed list of programming languages, since Node.js and Ruby on Rails are not languages, but frameworks.


LeaVerou:
It’s an example. It's more important to include varying enough strings than complete accuracy in terms of what constitutes a programming language.
If you are so concerned about this, you could send a PR to change the label instead.

PR #17082, opened 2017-07-12
Avoid script injections when rendering suggestions.

jcbrand:
Currently (before this change), any HTML contained in the suggestion text
will be rendered by `$.create` due to the fact that `innerHTML` is being used.

This means that if there are any scripts in the suggestion text, those
scripts will be executed, which is a security risk.

For example, if the suggestion text is ``
and the user types the text `<`, then you'll see an alert popup showing "123"
as that suggestion text gets inserted and rendered.

What this change does is to render as text (using `textContent`) the suggestion
text, while making sure that the `<mark>` elements are still rendered as HTML.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -322,11 +322,22 @@ _.SORT_BYLENGTH = function (a, b) {
 };
 
 _.ITEM = function (text, input) {
-	var html = input.trim() === "" ? text : text.replace(RegExp($.regExpEscape(input.trim()), "gi"), "<mark>$&</mark>");
-	return $.create("li", {
-		innerHTML: html,
-		"aria-selected": "false"
-	});
+    input = input.trim();
+    var element = document.createElement("li");
+    element.setAttribute("aria-selected", "false");
+
+    var regex = new RegExp("("+input+")", "ig");
+    var parts = input ? text.split(regex) : [text];
+    parts.forEach(function (txt) {
+        if (input && txt.match(regex)) {
+            var match = document.createElement("mark");
+            match.textContent = txt;
+            element.appendChild(match);
+        } else {
+            element.appendChild(document.createTextNode(txt));
+        }
+    });
+    return element;
 };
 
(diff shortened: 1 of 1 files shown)


jcbrand:
I'll investigate the test failures.

LeaVerou:
Hi there, thanks for contributing.

From a quick look, this looks like it completely disables HTML for suggestions and turns them all into text nodes? If that is true, it's throwing away the baby with the bathwater…

jcbrand:
Hi Lea

Yes, with this change the suggestion text is always rendered as text, not HTML.
However, the `<mark>` elements still render, so you see the matched parts.

Do you consider it a valid usecase for the suggestion text to be markup and for it to be rendered?

If you'd want the suggestion text to rendered as HTML, then to avoid executing scripts you'd need to santize the suggestion text/markup somehow to remove potentially malicious code (for example by stripping out `<script>` tags and their contents).

I'm however not sure this is a good idea.

If the suggestion text that is shown by awesomplete is derived from user-generated data, then an attacker could generate a malicious script (like the `img` tag shown above) that would be rendered by awesomplete as a suggestion, thereby allowing the code to execute on other users' computers.

If it was my site on which someone was doing this, I'd prefer seeing the markup as text in the suggestion results, so that I can see that something is fishy. Otherwise, if the markup was rendered as sanitized HTML, then the attackers could easily hide their tracks.

LeaVerou:
Awesomplete is already in use on a number of sites, many of which depend on HTML support. Even if it was starting today, HTML support is important. Sorry, but I’m gonna have to close this.

jcbrand:
That's unfortunate.

I need this patch for [converse.js](https://github.com/jcbrand/converse.js) so I pushed my fork to NPM.

For other people who might also be affected by this and are concerned about XSS attacks, the patched version is available here: https://www.npmjs.com/package/awesomplete-avoid-xss

PR #17083, opened 2017-07-14
Disallow multiple dropdowns

saintsGrad15:
To reproduce the error:

1. In a text box with awesomeplete enabled, choose an enumerated value.
2. Blur the text box
3. Focus the text box, displaying the drowdown.
4. All without the mouse, select some portion of the text (even all of it if minChars == 0) and delete the selected text
5. Observe that a second box appears but quickly covers the original box.
6. Click one of the options closing the box.
7. Observe that another box is still present.

Sometimes there are several although I can only reliably reproduce a two-box scenario.

See screenshot below. This was observed in:
FF 52.2.1
Chrome Canary 61.0.3150.0
Chrome 59.0.3071.115

See below for a screenshot of the developer console.

image

In addition a minor style typo was corrected and a Webstorm (Jetbrains IDE) hidden directory was added to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules/
 coverage/
+.idea/
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -122,7 +122,7 @@ _.prototype = {
 			this._list = list;
 		}
 		else if (typeof list === "string" && list.indexOf(",") > -1) {
-				this._list = list.split(/\s*,\s*/);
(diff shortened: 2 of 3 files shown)


LeaVerou:
Hey, thanks for contributing!

If calling `this.open()` is a problem when the dropdown is already open, shouldn't the `isOpened` check be in the `open()` method itself?

ghost:
Experiencing the same issue with Turbolinks and Ruby on Rails. Any update on this?

PR #17085, opened 2017-08-08
add possibility to use another container for suggestion items list (e…

bsvobodny:
add possibility to use another container for suggestion items list (ex: document.body)

Sorry for all the indentation/space diffs....


LeaVerou:
Hi there,

Thanks for contributing! This looks like a useful feature, but you'd need to fix the indentation issues before I can even review it...

PR #17093, opened 2017-10-02
Added option of removing highlighting from selection list

MaciejKus:
This adds an option "hideMark" which allows users to remove the yellow background highlighting that matches the user input.


LeaVerou:
Hey, thanks for contributing, but I'm not sure why have such an option. This is trivial to remove via CSS. E.g.

```css
.awesomplete mark { background: none !important }
```

PR #17098, opened 2017-11-09
Code from the example break labeled item list

GMartigny:
Hi,
A tiny change to maybe saves some time to someone.

If you copy-paste code from this example, it break the label/value pair item type.

### DIFF
--- a/index.html
+++ b/index.html
@@ -405,7 +405,7 @@ We can also <strong>generate list items based on user's input</strong>. See E-ma
 
 	replace: function(text) {
 		var before = this.input.value.match(/^.+,\s*|/)[0];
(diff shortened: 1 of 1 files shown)


LeaVerou:
Please take a look at [the API docs](http://leaverou.github.io/awesomplete/#extensibility)
`text` in the `replace()` function is supposed to be a string, not an input.
So, this change would actually break many more cases.

GMartigny:
I based this on the [REPLACE function](https://github.com/LeaVerou/awesomplete/blob/gh-pages/awesomplete.js#L352) of the code base which use `this.input.value = text.value;`

Issue #17113, opened 2018-01-10
Awesomplete does not cleanup completely


ferencbeutel4711:
Hi there,

I have noticed that after awesomplete is used, there is an element left at the bottom of the body which has the class 'visually-hidden'. Why is that? In my concrete case, it makes these elements visible, since your css is not loaded at all times on the page, the element persists however.

Thanks for looking into this.

LeaVerou:
This looks related to the a11y PRs that have been merged a few months ago.
@mauriciosoares @stephenkeable @ginader worked on that part of the code, so ideally, one of them could address this issue?
It should be a simple fix, probably just removing the node after the Awesomplete is done.
@ferencbeutel could you please paste more details about that element? What does its source look like in the inspector?

stephenkeable:
@ferencbeutel The element below that is visually-hidden, is a "status" which is used by screen readers, to give users that might be visually impaired additional information, such as the number of elements in the list. Or if they have not typed enough characters for the autocomplete to kick in yet.

In theory this could be shown to other users, although understand it's location would mean it would be hidden by the dropdown list possibly.

ferencbeutel4711:
Hi there,

sorry for the late feedback.
To further clarify the issue, here is the obfuscated html of the element in question:

```html
<div class="custom_awesomplete" style="left: 608px; top: 415px;">
  <ul hidden="" style="width: 464px;">
    <li aria-selected="true"><span>option1</span></li>
    <li aria-selected="false"><span>option2</span></li>
    <li aria-selected="false"><span>option3</span></li>
  </ul>
  <span class="visually-hidden" role="status" aria-live="assertive" aria-relevant="additions">option1</span>
</div>
```

This div node is inserted at the very bottom of the body.
After the user interacted with Awesomplete, the ul list is invisible. However, the span is not hidden and visible. As I saw in your css, you are styling visually-hidden elements, however, your css is not loaded anymore after the interaction with Awesomplete is done. So it would be very much appreciated if awesomplete would remove itself completely from the dom after a user interacted with it.

LeaVerou:
@stephenkeable My impression is that the status is retained in the DOM after Awesomplete is done.
I'm not an a11y expert, but I believe it's not needed anymore at that point?

stephenkeable:
Ah ok, think I understand fully now. After a user has made a choice it would announce which item was selected. After that it could be removed for sure, as would be added back on a new init of Awesomplete.

[...]

PR #17118, opened 2018-01-22
Fix for #17113

stephenkeable:
Added html hidden attributes to status element, on close and remove on
open.
Tested with Apple VoiceOver on Mac OS and status is still spoken
allowed.


LeaVerou:
Thanks for working on this!
Not sure how there are conflicts, since this is a fresh PR. 
It does look like there is a bunch of other changes though as well, could you take a look at the diff? Is that normal?

stephenkeable:
Forgot to merge the current branch into mine before making changes, rookie error.

LeaVerou:
Much better, thanks! 
I don't understand though, why apply `hidden` to it instead of removing it from the DOM and re-adding it on `open()`?

stephenkeable:
There is an instruction "Type x or more characters for results." given to user/screen reader, before open() is fired. 
Also matching the behaviour of the `<ul>` which is hidden, but left within the DOM.

LeaVerou:
Ah ok. Merged, thanks!

PR #17127, opened 2018-03-13
Optimize check li tag

Connormiha:
`tagName` is prefered for html elements and always in upper case. So it possible to avoid of regexp.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -103,7 +103,7 @@ var _ = function (input, o) {
 
 				if (li !== this) {
 
-					while (li && !/li/i.test(li.nodeName)) {
+					while (li && li.tagName !== 'LI') {
 						li = li.parentNode;
 					}
 
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -103,7 +103,7 @@ var _ = function (input, o) {
 
 				if (li !== this) {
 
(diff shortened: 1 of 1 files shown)


LeaVerou:
> tagName is prefered for html elements

[citation required]

Connormiha:
@LeaVerou 
https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName

`Returns the tag name of the element on which it's called. For example, if the element is an , its tagName property is "IMG" (for HTML documents; it may be cased differently for XML/XHTML documents).`

Connormiha:
nodeName property has all nodes includes comment, DocumentFragment etc. This is a more general property.
`tagName` only for Element. 
https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeName

LeaVerou:
I’m aware that tagName exists. I was aware even when it was non-standard, and nodeName was the only standard one.

I was asking for evidence that it's preferred or recommended.
Do note that tags will still be lowercase in XHTML, so this would be a breaking change. Therefore, even if we switch to tagName, we still need the case-insensitive matching.

Connormiha:
Yes, you right. In xhtml tagName in lower case.

PR #17129, opened 2018-03-18
Added an installation section

RandallFlagg:


LeaVerou:
Thanks for contributing!

Please remove the part about this being the "easiest" way, that's pretty subjective and makes non-programmers who typically don't use these tools feel unwelcome. It's easiest only for those who use yarn/npm for other things. For many others, especially less experienced folks, the easiest is to just include a script tag from a CDN. To be clear, I do not object to listing ways to install this via yarn or npm, what I object to is saying that this is the "easiest" way, or not including the script tag method.

Also, the npm line is about cookieconsent?!

RandallFlagg:
##
Here is my updated version.
Is this OK before I create another pull-request?
##

## Installation
There are a few ways to obtain the needed files.
Here are 2 of them:
1. CDN server

```sh
https://cdnjs.com/libraries/awesomplete
```

2. Another way to get up and running is by using `yarn` or `npm`:

```sh
yarn add awesomplete
```

```sh
npm install awesomplete --save
```

More information about the npm package can be found [here](https://www.npmjs.com/package/awesomplete).

LeaVerou:
Sure!

LeaVerou:
Merged, thanks!

Issue #17084, opened 2017-08-04
Awesomplete is not defined


nilsmielke:
I know, this has been discussed before, but none of the threads provided any help for my case.  
I can make Awesomplete work in its basic version using "data-list" on the input.  
As soon as I try to instantiate in JS, I get the error  
`Uncaught ReferenceError: Awesomplete is not defined at kunden-liste?filter=active:79`  
in the console.  
My input field looks like this:  
`<input placeholder="suchen" id="firma">`
JS:  
```
<script>
var input = document.getElementById("firma");
var awesomplete = new Awesomplete(input);
// Show label but insert value into the input:
awesomplete.list = [
        { label: "Belarus", value: "BY" },
        { label: "China", value: "CN" },
        { label: "United States", value: "US" }
    ];
</script>
```
Any clues are appreciated.

LeaVerou:
Sounds like Awesomplete is not available yet. Possible causes:

1. The script that includes it is after this one
2. The script that includes has `async` and/or `defer`.

delibes:
@LeaVerou it might be a bit more of a timing issue than that.

I'm getting a similar problem. FWIW, I am bundling this library with other code using uglify, but I don't think that changes the behaviour much.

The private _ variable is exposed to the global scope with ```self.Awesomplete = _;```

Unfortunately the init() call can take please before this happens, and then the internal function tries to reference the undefined global in ```Awesomplete.count = (Awesomplete.count || 0) + 1;```

Moving  the ```if (typeof self !== "undefined")``` block a few lines up before the ```if (typeof Document !== "undefined")``` block seems to ensure everything is initialised when needed.

I'll try submitting a pull request when I get a chance.

PR #17131, opened 2018-03-28
Ensure global Awesomplete on self is defined

delibes:
Fix for #17084 

Ensure global Awesomplete on self is defined before any possible call to init(), so that `Awesomplete.count` doesn't break.

I did look at adding a Karma/Jasmine test for this, but "it's hard"!

I would have to start changing all the test setup, which might not be a popular choice. I think a part of the problem is that `karma.conf.js` line 12 ensures that Awesomplete is always loaded early and thus initialised via the DOMContentLoaded event.

That means this code is never actually run in any test:
```
	if (document.readyState !== "loading") {
		init();
	}
```

Perhaps somebody with more Karma experience might be able to offer a better way to control the loading?

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -498,6 +498,11 @@ function init() {
 	});
 }
 
+// Make sure to export Awesomplete on self when in a browser
+if (typeof self !== "undefined") {
+	self.Awesomplete = _;
+}
+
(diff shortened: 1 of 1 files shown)


LeaVerou:
Merged, thanks!

PR #17135, opened 2018-04-12
Ajax support added

Beingless:
A an Ajax callback to fetch list from server.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -9,7 +9,7 @@
 
 var _ = function (input, o) {
 	var me = this;
-    
+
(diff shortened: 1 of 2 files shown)


LeaVerou:
Hi there,

Thanks for contributing. However, I'm afraid this is out of scope. The API already permits this, and we'd rather keep Awesomplete small.
I'd welcome a docs example, but I don't think changes in the core are needed.
Furthermore, I'm not sure how this makes it any easier, it looks like the example still requires most of the Ajax logic to be written every time.

PR #17138, opened 2018-05-15
Autofocus support

nxtpl:
Autofocus was not working, the fix resolves this problem.


LeaVerou (changes requested):
Hey, thanks for this fix!!

Why duplicate the appendChild instead of just:

```js
if (ref.getAttribute("autofocus") != null) {
	ref.focus();
}
```
?

nxtpl:
Good point, I will submit new PR in a moment :)

nxtpl:
Updated the code

LeaVerou:
LGTM, merged, thanks!

PR #17142, opened 2018-05-27
container function to generate container element

melitele:
I've added a `container` function to options that will allow client to provide a different container element. The default version of `container` function creates container element as before. I believe this is in line with the `item` function that customizes creating item elements.

This addresses my use case where container element cannot reuse the input element parent
because the input parent clips container's content. It also addresses a case when input element has to be next to it's original siblings.

It also solves #17086 & #16718

Thank you for consideration.


melitele:
I just noticed a bug - I'll fix it and resubmit the PR. My apologies.

PR #17143, opened 2018-05-27
container function to generate container element

melitele:
I've added a container function to options that will allow client to provide a different container element. The default version of container function creates container element as before. I believe this is in line with the item function that customizes creating item elements.

This addresses my use case where container element cannot reuse the input element parent
because the input parent clips container's content. It also addresses a case when input element has to be next to it's original siblings.

It also solves #17086 & #16718

Thank you for consideration.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -23,7 +23,9 @@ var _ = function (input, o) {
 	this.input.setAttribute("aria-owns", "awesomplete_list_" + this.count);
 	this.input.setAttribute("role", "combobox");
 
-	o = o || {};
+	// store constructor options in case we need to distinguish
+	// between default and customized behavior later on
+	this.options = o = o || {};
 
 	configure(this, {
 		minChars: 2,
@@ -32,6 +34,7 @@ var _ = function (input, o) {
 		data: _.DATA,
 		filter: _.FILTER_CONTAINS,
 		sort: o.sort === false ? false : _.SORT_BYLENGTH,
+		container: _.CONTAINER,
 		item: _.ITEM,
 		replace: _.REPLACE
 	}, o);
@@ -40,10 +43,7 @@ var _ = function (input, o) {
 
 	// Create necessary elements
 
-	this.container = $.create("div", {
-		className: "awesomplete",
-		around: input
-	});
+	this.container = this.container(input);
 
 	this.ul = $.create("ul", {
 		hidden: "hidden",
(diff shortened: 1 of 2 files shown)


awesomplete.js
```diff
@@ -40,10 +41,7 @@ var _ = function (input, o) {
 
 	// Create necessary elements
 
-	this.container = $.create("div", {
-		className: "awesomplete",
-		around: input
-	});
+	this.container = this.container(input);
```
LeaVerou, line 46: Do note that this means the `this.container()` function is overwritten and not accessible anymore. Are we sure we don't need to access it again?

melitele: If I am reading the original code correctly, the container element is never re-created during the object life cycle. That's why I've let the function to be overwritten. I can change the field used for the container function - do you have a preference for the field name?

Thank you for looking at this PR.

awesomplete.js
```diff
@@ -203,11 +201,13 @@ _.prototype = {
 		$.unbind(this.input, this._events.input);
 		$.unbind(this.input.form, this._events.form);
 
-		//move the input out of the awesomplete container and remove the container and its children
-		var parentNode = this.container.parentNode;
+		if (this.input.parentNode === this.container) {
```
LeaVerou: What's the reasoning behind this if?

LeaVerou (commented):
Hey, thanks! Looks good overall, please reply to these 2 comments.

awesomplete.js
melitele: The container element should only be destroyed by the `Awesomplete` if it was created by the `Awesomplete` - the condition of the if is `true` when the container element was created by the default method of creating it.

LeaVerou: Ok, fair enough.

awesomplete.js
LeaVerou: Nah, I think `container` is a good name. Perhaps we can store the constructor options somewhere? Like `this.options`? Then we still have access to the function, so you can compare it to the default function in the check in `destroy()` instead of the less foolproof way you're checking now.

awesomplete.js
LeaVerou: But see my other comment about improving the check.

LeaVerou: Basically, what do we want if the container *is* custom, but also a parent of the input? Do we want to destroy it or not? That affects what we check for here.

awesomplete.js
melitele: Done. Added `this.options` to hold the original options passed to `Awesomplete` constructor. Checking `this.options.container` in `destroy()` and cleaning up the container element only when `this.options.container` is not set i.e. when container is originally created by `Awesomplete`.

LeaVerou:
Nice, merged! Thank you so much!

melitele:
Thank you for taking my PR. And thank you for making and sharing this component.

PR #17144, opened 2018-05-30
change reference to itself to be _ as everywhere else

melitele:
Awesomplete should refer to itself as `_` instead of `Awesomplete`
because the latter is only globally available in the browser

While this is mostly to be consistent - everywhere else `_` is used internally - the issue happens to affects tests that I run outside browser.

Thank you for consideration.

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -11,8 +11,8 @@ var _ = function (input, o) {
 	var me = this;
     
     // Keep track of number of instances for unique IDs
-    Awesomplete.count = (Awesomplete.count || 0) + 1;
-    this.count = Awesomplete.count;
+    _.count = (_.count || 0) + 1;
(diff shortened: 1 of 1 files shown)


LeaVerou:
Wow, you're on a roll :)
Looks good to me, have you tested it?

melitele:
I ran karma Awesomplete tests that passed (including the one that tests the generated id based on `this.count`). I have also tested the change in my own module that uses Awesomplete - both in the browser and out of it (the reason for the fix).
I'll run any additional tests you suggest.

LeaVerou:
Nah, that's what I wanted to hear so I could avoid doing it myself :)
Merged, thanks!

melitele:
Thank you for taking the patch.

PR #17145, opened 2018-06-05
Implement tab to select as option defaulting to false

kball:
Targeted fix for #16832 behind an option. Allows specifying if you want the 'tab' key to select current option or not.


LeaVerou:
Looks good, merged, thanks!

Note to self: at some point we should switch to checking `evt.key` instead of these archaic keycodes...

Issue #17155, opened 2018-08-24
autocomplete="off" doesn't work in Chrome


FranciscoG:
using `autocomplete="off"` does not stop Chrome's autofill from working.  

The current workaround is to pass a random value to it

I use `autocomplete="nope"` and it works great.  You can put in any string value that is not "off" or "on" in there.  

I suggest `autocomplete="awesomplete"`

https://www.20spokes.com/blog/what-to-do-when-chrome-ignores-autocomplete-off-on-your-form

https://stackoverflow.com/a/38961567/395414

[link to the spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill)

LeaVerou:
Thanks! PR?

antoinelyset:
BTW I thought my PR would fix this but for the new Chrome's version, it didn't. 😭

antoinelyset:
Ok, after some investigation, it seems to fix some cases and some not. Maybe you can try in your own context @FranciscoG.

blvrd:
@antoinelyset your PR seems to have broken the autocomplete behavior in Firefox 63. I'm now seeing the native autcomplete menu on top of the Awesomplete menu:


![screen recording 2018-09-21 at 01 38 pm](https://user-images.githubusercontent.com/3522323/45899890-c3f91500-bda3-11e8-9e97-ced3474bbcd3.gif)

LeaVerou:
Ugh, so I guess if it didn't fix it in Chrome and it broke it in Firefox, I'm gonna revert it until we figure out the best course of action.

blvrd:
@LeaVerou thanks so much for the quick turnaround.

LeaVerou:
No worries, though I did a release yesterday, and now I'm not sure what to do. :(

antoinelyset:
Aargh. This is problematic indeed, no good solution...

For the record it did fix the Chrome's bug in pre-69 version. Since 69 it's back.

We should let it `"off"`, but maybe being able to parametreize would be useful.

For the release, just bump it to 1.1.4, it will like 1.1.2 with some fixes included.

FranciscoG:
Sorry about this!  Very frustrating indeed that Chrome is inconsistent.  My original solution worked for my needs at the time but I have since switched jobs and have not been able to test it since.

adamgruber:
Is there an anticipated timeframe for releasing the reverted change made in https://github.com/LeaVerou/awesomplete/pull/17162?

[...]

PR #17159, opened 2018-09-14
Add "autocomplete" as initialization option

antoinelyset:
Hi 👋 !

I have the same issue as #17155, but I suggest a slightly different modification. I don't know if this is a priority but I feared that by modifying the default `"autocomplete"` value we may break some users' usage (even if this is an undocumented behaviour). Moreover by defaulting to `"off"` we are still following the [best practices](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill).

What do you think ?

And one more question, do I need to update the built/generated files ?

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -19,7 +19,7 @@ var _ = function (input, o) {
 	this.isOpened = false;
 
 	this.input = $(input);
-	this.input.setAttribute("autocomplete", "off");
+	this.input.setAttribute("autocomplete", "awesomplete");
 	this.input.setAttribute("aria-owns", "awesomplete_list_" + this.count);
 	this.input.setAttribute("role", "combobox");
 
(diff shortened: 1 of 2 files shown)


LeaVerou:
Hey, thanks for submitting a PR!

I wouldn't worry about breaking usage. There doesn't seem to be any use case for having **both** Awesomplete and the browser's autocomplete.
With this, people would need to specify the required value manually, when all they want is to disable the native autocomplete. This is against the UX best practice of sensible defaults.

antoinelyset:
I understand, should I still keep the possibility to pass another value as an option ?

LeaVerou:
It seems a little premature to me. We can always add it very easily if even one person requests it, but having a feature with zero use cases seems excessive.
Hope that makes sense!

antoinelyset:
I completely understand (I wanted to set it to "new-address" for a more semantic name).
I made the change.

LeaVerou:
Merged, thanks!

PR #17173, opened 2019-01-23
fix a couple accessibility issues

triptate:
* list items under a ul with role listbox need role option
* the role combobox needs an aria-expanded attribute that tracks the state of the combobox


LeaVerou:
Thank you!

Issue #17079, opened 2017-06-20
Grabbing the selected value from a label/value list?


wildsky:
So if I instantiate with a list like this:

`awesomplete.list = [
			[ "Belarus", "BY" ],
			[ "China", "CN" ],
			[ "United States", "US" ]
		];`

Then I'll end up with the value in my user's input. I'm trying to have the label appear in the input upon selection, but also be able to know what value matches their choice, so I can stick that in a hidden input.

So for the above, if the user chose "China", I'd like "China" to show up in their choice, but have "CN" get stashed in a variable/hidden input field.

Is Replace the right method for doing this?

zampage:
just instantiate the list like this:
`awesomplete.list = [ {label: "Belarus",  value: "BY"}, {label: "China", value: "CN"}, {label; "United States", value: "US"} ];`

If you want to do something upon selection (eg. write value/label to a hidden field) check the event `awesomplete-selectcomplete` and use the `event.text.value` / `event.text.label` variables.

alexcittadini:
I'd also like to see this resolved.

cutme:
`event.target.value`

briandipalma:
This is baffling to me. Why does awesomeplete put the `value` in the input? Surely the whole point of having labels and values are to show/provide the `label` to the user as user friendly piece of text while the `value` is the machine/data store friendly data. For example having i18n text in the label but using `value` as the key to the data in a store/back end system.

@LeaVerou What am I missing here?

briandipalma:
OK, I found this other issue, https://github.com/LeaVerou/awesomplete/issues/17104 it looks like maybe adding this to the documentation would help. I can create a PR but unsure if it will be merged given the backlog?

LeaVerou:
> I can create a PR but unsure if it will be merged given the backlog?

PRs are not merged FIFO. Easy, small, uncontested PRs are merged ahead of any backlog.

briandipalma:
Cool, I added a bit to the documentation explaining this. I was almost going to link the GH issue but that feels a bit messy. I'd probably close this issue once the PR is merged as it addresses the question. I'm still doubtful of the choice on value over label but changing that is a breaking change.

PR #17178, opened 2019-03-08
docs: Explain how to insert label into input

briandipalma:


briandipalma:
Should close #17079

LeaVerou:
Merged, thanks!

I might try to be a bit clearer that using a custom replace function is a choice depending on what you need inserted, not something they *need* to do.

PR #17180, opened 2019-03-14
feat: Add `originalEvent` property to select events

### LINKED ISSUE #17179: Distinguish between click and enter selections
The event `"awesomplete-selectcomplete"` provides you with the selected option but it doesn't tell you the user action that triggered the selection. Could I create a PR that provided the underlying event?

Basically `keydown` would pass the `evt` value to `select`

`me.select(undefined, undefined, evt);` and `select` would pass that on

```
"awesomplete-selectcomplete", {
  evt: evt,
  text: suggestion
}
```

and similar for `click`. Would that be an acceptable change?

briandipalma:
Allows distinguishing between click and keydown selection.

Should close #17179

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -77,10 +77,10 @@ var _ = function (input, o) {
 				if(me.opened) {
 					if (c === 13 && me.selected) { // Enter
 						evt.preventDefault();
-						me.select();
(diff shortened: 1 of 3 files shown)


LeaVerou:
Thanks, merged!

PR #17204, opened 2019-09-24
Add documentation: TabSelect property

Robbertdk:
In commit da06e45f829a3d5c0323a9d0d9d5ff2def4825b5 the functionality to select the first item on tab is added, but this can't be found in the documentation. This change adds that documentation.

### DIFF
--- a/index.html
+++ b/index.html
@@ -220,6 +220,13 @@ new Awesomplete(input, {
 				<td>Boolean</td>
 				<td>false</td>
 			</tr>
+			<tr>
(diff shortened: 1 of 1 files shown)


LeaVerou:
Oof, We generally avoid merging PRs that add features without adding the corresponding documentation, but I guess this slipped through the cracks.
Merged, thank you!

onecreative:
Did this ever happen? I don't see it in the docs, and the feature doesn't appear to work.

PR #17206, opened 2019-11-04
Update documentation on behaviour of `destroy()`

kraftner:
Update documentation of `destroy()` to clarify that the container is only removed if it wasn't manually set via the `container` property.

This brings the documentation in line with the [inline code comment](https://github.com/LeaVerou/awesomplete/blob/12c913ba7ffbcbe033f580694fd5a7c996f4d001/awesomplete.js#L213-L220).


LeaVerou:
Merged, thanks!

PR #17226, opened 2020-09-16
docs: Fix simple typo, mousedowm -> mousedown

timgates42:
There is a small typo in awesomplete.js.

Should read `mousedown` rather than `mousedowm`.


LeaVerou:
Thanks!

PR #17236, opened 2021-04-26
Add ignore diacritics capabilities 

### LINKED ISSUE #17187: Highlighting word with accent
Hi..

Im trying to search accents words with a filter and work fine to me..But i need some help to highlight..

`var comboplete = new Awesomplete("dstAutoComplete", {
    minChars: 2, filter: function(text, input) {
      return Awesomplete.FILTER_STARTSWITH(text.normalize("NFD").replace(/[\u0300-\u036f]/g, ""), input.match(/[^,]*$/)[0]);
    },
  });`

In this example word **So** and **Só** must be highlight.

Teste Autocomplete

Sorry my english.

### LINKED ISSUE #17234: Show list of accented words without having to put the accent
I am wanting to implement Awesomplete to put on my website. I have been able to fetch a list of terms from the MySQL database. The problem is in the use of **accents in certain words**, which happens with Spanish. In this case I would like that the accent does not have to be written to match. For example, when searching for ""auditoría, programación, vínculo, etc.", they are only shown if I put the accent and what interests me is that they are shown without having to write the accent in its corresponding vowel.
For example, when I write "auditoria" (without accent) it shows me in the list.

scoopandrun:
# Description

Added the possibility to ignore characters with diacritics.
When the user inputs a string, it should match with the items in the list whether or not the input or the items have accented characters.

Fixes #17187 
Fixes #17234 

## Type of changes

- [x] New feature (non-breaking change that adds functionality)

This PR introduces a new option **Awesomplete.ignoreDiacritics** (boolean, default *true*).
When set to *true*, it allows matching accented characters against non-accented characters.
When set to *false*, it requires characters to be the same to match.

This is an implementation of [David Calhoun's blog post](https://www.davidbcalhoun.com/2019/matching-accented-strings-in-javascript/).

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -376,13 +376,14 @@ _.CONTAINER = function (input) {
 }
 
 _.ITEM = function (text, input, item_id) {
-	inupt = this.ignoreDiacritics ? removeDiacritics(input) : input;
+	input = this.ignoreDiacritics ? removeDiacritics(input) : input;
 	var text_comp = this.ignoreDiacritics ? removeDiacritics(text) : text;
 
 	// shadow = text without diacritics (for comparison)
 	var shadow = input.trim() === "" ? text : text_comp.replace(RegExp($.regExpEscape(input.trim()), "gi"), "<mark>$&<mark>");
 	
 	// Display diacritics in suggestions (html = with diacritics)
+	let html = "";
 	if (shadow.indexOf("<mark>") !== -1) {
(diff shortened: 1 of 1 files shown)


LeaVerou:
Hi there,

Thank you for working on this.
I think this should be a usage example, not part of the API surface. Hopefully the existing parameters allow for implementing this that way.

Later down the line we should have plugins, and this could be a plugin.

PR #17238, opened 2021-06-14
Prevent Tab key resulting in list element being skipped when `tabSelect: true`

AlexisDeschamps:
It seems the branch for handling a `tab` key event if `tabSelect: true` is missing an `evt.preventDefault()`, which results in an item being skipped when autocompletes are created in a list.

To fix this, I believe we could call `evt.preventDefault()` in all branches of the procedure, instead of only some.

I have manually tested that this fixes the problem on our UI. The existing tests also appear to pass. If this change seems to make sense, I can look into adding a test.

### Example
**Initial state**
image

**(Current behavior): final state after pressing tab:**
image

**(Expected behavior): final state after pressing tab:**
image

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -72,22 +72,24 @@ var _ = function (input, o) {
 			"input": this.evaluate.bind(this),
 			"blur": this.close.bind(this, { reason: "blur" }),
 			"keydown": function(evt) {
-				evt.preventDefault();
 				var c = evt.keyCode;
 
 				// If the dropdown `ul` is in view, then act on keydown for the following keys:
 				// Enter / Esc / Up / Down
 				if(me.opened) {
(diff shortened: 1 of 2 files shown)


LeaVerou (changes requested):
I would be wary about cancelling the event in every case, whether it's relevant or not. That could have serious accessibility consequences. For example, it disables tabbing even when `tabSelect` is not on, not to mention it cancels the event regardless of whether Tab was even pressed at all.
I'd welcome a PR that adds it under a more specific clause.

Thanks for contributing!

AlexisDeschamps:
@LeaVerou Thank you for the guidance. That makes sense. I modified the `preventDefault` call to be under a specific clause and added a unit test. Please let me know if there is something else I should add or change.

LeaVerou:
Merged, thanks!

Issue #17242, opened 2021-08-02
i18n support


bertolami:
Would be great to have a possibilty to translate the screen reader message, e.g. "list item 1 of 10", or "no results found" in order to support users that don't speak English

LeaVerou:
Agreed! I'd welcome a PR for this.

PR #17243, opened 2021-08-02
i18n support

bertolami:
#17242 

I tried to add simple i18n support. is currently running as part of our timetable: https://preview-timetable.dsiag.ch

Cheers, Roman

### DIFF
--- a/awesomplete.js
+++ b/awesomplete.js
@@ -39,7 +39,8 @@ var _ = function (input, o) {
 		item: _.ITEM,
 		replace: _.REPLACE,
 		tabSelect: false,
-		listLabel: "Results List"
+		listLabel: "Results List",
+		tNoResults: "No results found"
 	}, o);
 
 	this.index = -1;
@@ -328,7 +329,7 @@ _.prototype = {
 
 			if (this.ul.children.length === 0) {
 
-                this.status.textContent = "No results found";
+                this.status.textContent = this.tNoResults; // "No results found";
 
 				this.close({ reason: "nomatches" });
 
@@ -341,7 +342,7 @@ _.prototype = {
 		else {
 			this.close({ reason: "nomatches" });
 
-                this.status.textContent = "No results found";
+                this.status.textContent = this.tNoResults; // "No results found";
 		}
 	}
 };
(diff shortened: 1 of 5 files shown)


awesomplete.js
```diff
@@ -39,7 +39,12 @@ var _ = function (input, o) {
 		item: _.ITEM,
 		replace: _.REPLACE,
 		tabSelect: false,
-		listLabel: "Results List"
+		listLabel: "Results List",
+		tStatusResult: "${length} results found",
+		tNoResults: "No results found",
+		tStatusQueryTooShort: "Type ${minChars} or more characters for results.",
+		tStatusStartTyping: "Begin typing for results.",
+		tListItemText: "list item ${index} of ${length}"
```
LeaVerou, line 45: I think these should be in a separate object, not just as raw properties. Also, we do not need them on every instance, what we need is a way to select a language for the instance. So I'd suggest a data structure like: 

```js
Awesomplete.languages = {
	"en-us": {
		"list-label": "Results List"
	}
}
```

Then instances would point to one of these objects either by storing the language, or a reference to the object (I'll leave that up to you).

awesomplete.js
```diff
@@ -62,7 +67,7 @@ var _ = function (input, o) {
 		"aria-live": "assertive",
         "aria-atomic": true,
         inside: this.container,
-        textContent: this.minChars != 0 ? ("Type " + this.minChars + " or more characters for results.") : "Begin typing for results."
+        textContent: this.minChars != 0 ? this.tStatusQueryTooShort.replace("${minChars}", this.minChars) : this.tStatusStartTyping
```
LeaVerou: We probably need a method for this instead of just raw `.replace()` calls (also note that if the phrase needs to continue the substitution multiple times, this will only substitute the first one).

For inspiration, check out how Mavo does this: https://github.com/mavoweb/mavo/blob/master/src/locale.js#L25

LeaVerou (changes requested):
Hey, thanks for working on this! I left some comments.

bertolami:
Hello @LeaVerou 
Could you please check again. I added a static field with translations and a field with the language. If it makes sense to you now, I would update the doc.
Cheers, Roman

Issue #17252, opened 2024-03-30
On Android in Talkbalk (screenreader) it is not possible to select a suggestion from the list


nico-amsterdam:
Tested this on a Android phone width Talkback  14.1 (currently the latest) on Chrome .
I can type something in the input field, but if I want to select something from the suggestions it selects the whole UL listbox, and it is not possible to select one item.
Turns out the the LI items in the suggestion list must be put in the tab sequence.
If you add the following in the _ITEM function in awesomplete.js, below line 381, it works fine:
   "tabindex": "0",

Although Awesomplete works great, the whole project could use a bit of LOVE.
- 95% of the issues can be closed with the remark: 'functions as designed'   or   'closed due to 2 years of inactivity'   or 'won't fix'  or 'workaround can be made with a little bit of effort'. Currently the project looks abandoned.
- The listLabel text can be configured, but the other a11y text (status.textContent) not. This would be useful if you want tot use Awesomplete in a different language with a screenreader. Not every blind person understands English, if you pardon my French :-)
- And a minor bug: in the destroy function, the attribute 'aria-autocomplete' is removed (line 227), but this attribute is never set by Awesomplete. Looks like it is missing around line 22. And shouldn't the other attributes (aria-owns for example) be removed in the destroy function as well?

TIA,

      Nico

nico-amsterdam:
PR is merged, so this ticket can be closed.
Thanks

PR #17254, opened 2024-04-17
fixes issues of #17252. a11y improvements. fix Talkback item selection. i18n for the status text.

nico-amsterdam:
- add tabindex to fix the issue that it is not possible to select an item when using the Talkback screen reader
- removed the screen reader text that is not used by Voiceover and Talkback, and superfluous for NVDA and Narrator
- new config options to pass on a different text for the status used for screen readers, enabling the possibility to supply a text in a different language
- update package.json to fix gulp
- fix leading tab characters in the source code (although personally I prefer all spaces)
- update tests
- update minimized version of awesomplete


LeaVerou (approved):
Looks great, thank you so much!

Issue #17255, opened 2024-06-26
Tab sequence when suggestions are shown; need to tab twice to tab to the next field


nico-amsterdam:
Although my PR https://github.com/LeaVerou/awesomplete/pull/17254 fixes the issue that in the Talkback screen reader you couldn't select anything from the suggestions, it also introduced a minor annoyance in the tab sequence:
- if suggestions are shown and you hit the tab key in the input field to leave the field, it doesn't go immediately to the next field in the tab sequence. It tries to go to the suggestion first. You have to enter the tab key another time to get in the next field.

Luckily there is an easy fix.
Instead of having tabindex 0 on the list items, it is enough to set tabindex -1 to make the list items focusable elements.
I tested this in the Talkback screen reader, and this works just as good and this doesn't disturb the tab sequence.

I will submit a PR for this.
I was thinking about adding a test, but as far as I know, it is not possible to test the tab sequence with Jasmine tests.

In the PR I will also make aria-activedescendant empty when no suggestion is selected. Currently aria-activedescendant keeps pointing to a list item even though the popup with items are hidden. This doesn't seem to cause any issues, but it is cleaner to make aria-activedescendant empty.

nico-amsterdam:
PR with the fix is merged. Closes this ticket.

PR #17256, opened 2024-06-26
Fixes issues of #17255. Fix tab sequence

nico-amsterdam:
awesomplete.js:
- Use tabindex -1 instead of 0 to make elements focusable. This does not disturb the tab sequence. Tested with the Talkback screen reader.
- Make aria-activedescendant empty when none is selected. This is when 
  - none of the items has aria-selected=true
  - or the popup is closed 

awesomplete.theme.css
- After the transition of 0.3 seconds the visibility should be hidden. Prevents navigation to anything inside the now hidden popup.

awesomplete.css is updated by gulp because of the change in awesomplete.theme.css


dpb-bah:
Confirmed this works. We've been using awesomplete for a while but it flags issues with AXE scans for Section 508 compliance. This PR resolves it.

LeaVerou (approved):

LeaVerou:
Thanks for checking it out @dpb-bah and for contributing @nico-amsterdam!

